PR #2, opened 2023-09-08
Add exceptions to calls to junit fail

19mdavenport:
Adding the exception into the call to `fail` makes it a bit easier to debug problems causing said exception


leesjensen (approved):
Looks great! Thanks

PR #4, opened 2023-09-09
Add example value for field "gameID"

pawlh:
All other instances of `"gameID"` in the specs give an integer example value, this proposed change adds an example to an instance missing a value


leesjensen (approved):
Thanks!

PR #9, opened 2023-09-18
Update specification.md

maypeter567:
inconsistent use of one when 1 is used more


leesjensen:
Thanks for the improvements. I went the other way and spelled everything out for consistency.

PR #10, opened 2023-09-25
Update ChessBoardTests.java

prattnj:
## New Features
Add tests to check that `ChessBoard.resetBoard()` also resets the middle squares to null.


leesjensen:
Looks great! Thanks.

PR #13, opened 2023-09-28
Lint markdown, minor fixes and clarifications

ThanGerlek:
Biggest change is simply linting all the markdown files using VSCode's [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) extension.

I also added links between related markdown files for the chess project, but someone else did almost the same thing and PR'd before I did, which gave me one heck of a merge conflict. (And forced me to finally learn how to use remotes from the git CLI.) But I can claim the emojis at least.


leesjensen:
Nathaniel, thanks for taking the time to do this work. There are some great corrections in here, but it looks like the content moved out from underneath you between when you created you commit and the changes you are suggesting. For example, the changes to `Design Principles` seem to delete all of the content that to current version contains. Perhaps this is a result of the merge changes you described.

I'm not sure how to easily resolve this without either losing some important content, or losing your changes. I'm mostly interested in your corrections that fix things that are busted, like typos and missing links. The markdown linter is less crucial since the rendering fixes most of those problems anyway.

What if you started will a clean copy of the current code and modify the essential changes with a new pull request?

PR #16, opened 2023-10-02
fix ChessBoardTests addGetPiece

19mdavenport:
The add and get piece test in ChessBoardTests used the same ChessPosition for both add and get. I've seen students pass this test and assume everything is well and then get really confused why resetBoard fails because of their equals and/or hashcode not working. Using a different but equals object should fix this


leesjensen (approved):
Great addition!

PR #18, opened 2023-10-03
add tests for equals and hashCode

19mdavenport:
Not sure if you want to add any or all of these, but here are some ideas I had for testing equals and hashCode for ChessPosition and ChessMove


leesjensen (approved):
Looks great! Thanks.

PR #19, opened 2023-10-05
remove 'Override Methods' from the instructions

pawlh:
The instructions for overriding `equals` and `hashCode` currently mention selecting "Override Methods" from the Code -> Generate menu, but this just creates code that  calls the super methods.

This change removes mention of "Override Methods" altogether and corrects which sub-menu choice should be selected


pawlh:
I am running IntelliJ on a Linux device, so it is possible the menu choices are different on other platforms, but I doubt that

pawlh:
It looks like `Code -> Generate -> equals() and hashCode()` only appears if a class already has fields, so students might get confused if they try to generate these methods before finishing implementing the classes.

I added a note about this to my PR, but if there is a better way to explain this I welcome changes!

PR #20, opened 2023-10-15
Fix typo in List Games success response

CDJ64:
To be honest, I'm not familiar with this format (JSON?), but I'm guessing there's a typo since there's only one curly brace on the left and two on the right. If I inserted it incorrectly, please adjust accordingly.


leesjensen (approved):
Thanks for the correction! You got it right.

PR #21, opened 2023-10-17
Update unit-testing.md

CDJ64:


leesjensen (commented):
Is there a grammatical problem here, or is this just a stylistic preference?

CDJ64:
It's not so much a grammatical error as a sentence that doesn't make much
sense. What does it mean to say a given assertion function represents
"assertEquals"? That's just its name. I guess the name represents what the
function does, but it still doesn't make much sense. Another possibility
could be to say "the assertion functions do things such as..." but even
then it's kind of confusing.

PR #23, opened 2023-10-20
Moved topics between modules to match Canvas

CDJ64:
If Canvas should be changed instead, you can do that.


leesjensen:
I'll move it in Canvas. Thanks!

PR #24, opened 2023-10-20
Update web-api.md

dbowie83:
add link to example code


leesjensen:
@dbowie83, thanks for the suggestion. I changed it a bit to use a relative link so that we can be a little less coupled to the repository.

dbowie83:
Awesome thanks!

PR #25, opened 2023-10-23
Update instructions for passing of service class test cases

pawlh:
We decided to include grading the unit tests as part of the live pass off, so this aligns the pass off instructions with that.

Note, my understanding is that the automated test cases and the test web page are required before receiving any points for any part of the rubric (so pass/fail), but the student-made test cases are not. This would have the following implications (non-exhaustive):

1. A student passes off all automated tests and the test web page, but did not write any service unit tests
    a. They would get 125/150 for the live portion of the pass off
    b. They would not be able to pass off again, even if they went back and wrote service unit tests
2. A student fails some automated tests and fails the test web page and had not written any service unit tests
    a. They would not pass, but would be able to rejoin the help queue later to pass off again
    b. In this time, they could choose to write their service class unit tests

If my understanding is correct, then please review the new wording to make sure I expressed it clearly

### DIFF
--- a/chess/3-web-api/web-api.md
+++ b/chess/3-web-api/web-api.md
@@ -271,7 +271,7 @@ For this phase the TAs will grade the quality of your project's source code. The
 
 To pass off this assignment, meet with a TA and demonstrate that your code passes the provided test cases and that your test web page loads correctly. You must pass this part to recieve credit for any part of the assignment.
 
-During the pass off, the TA will check the test cases your wrote for the Services classes pass and grade them accoridng to the rubric.
+After checking the above, the TA will run and review the test cases your wrote for the Services classes pass and grade them accoridng to the rubric.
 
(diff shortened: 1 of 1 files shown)


leesjensen (approved):
This looks like a good clarification. Thanks!

leesjensen:
@pawlh, I'm not sure why you came to your conclusions about pass/fail and retakes. I'm guessing you are basing this on the retake policy for phase #2. For phase #3 I would just follow the rubric for awarding points.

pawlh:
> @pawlh, I'm not sure why you came to your conclusions about pass/fail and retakes. I'm guessing you are basing this on the retake policy for phase #2. For phase #3 I would just follow the rubric for awarding points.

@leesjensen This is based on the original instructions for Phase 3. We usually allow unlimited pass off attempts for the live pass off, but once you submit your code you can't resubmit for additional points (with Phase 2 being the exception to the rule). My conclusions about the pass/fail and retakes are based on the understanding that although we changed _when_  tests are graded, we haven't changed anything else (ie, unit tests still aren't going to be allowed to have unlimited attempts)

I left some additional notes about that in the grading guide in the TA channel

PR #29, opened 2023-11-01
Small typo in phase 4 getting-started.md

CDJ64:
"MYSQL_URL" might also need to be updated since my starter code gives it a different name


leesjensen (approved):
Thanks! I did a bit more to clarify how you use the Database class, but I appreciate you pointing out the inconsistency.

PR #32, opened 2023-11-07
Small typoes in console-ui.md

CDJ64:
Also, the Coloring With Java subsection's example code doesn't use constants like the text suggests.


leesjensen (approved):
Thanks!

leesjensen:
I also changed the text for the coloring to remove the reference to constants.

PR #33, opened 2023-11-10
fix(jdbc) code typo

19mdavenport:
https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/JsonDeserializationContext.html


leesjensen (approved):
Blah! Thats what I get for not verifying the code. Thanks for watching my back. Get suggestion to use the ctx.

PR #41, opened 2024-01-09
use echo for git example

pawlh:
`printf` only works on Mac and Linux, but `echo` should work on Windows as well


leesjensen (approved):
Thanks!

PR #42, opened 2024-01-11
Replace Outdated Setup Instructions

frozenfrank:
Old instructions from last semester were instructing students to clone the entire repo. Those instructions have been replaced with a reference to the new instructions which live on a different page.

### DIFF
--- a/instruction/java-fundamentals/java-fundamentals.md
+++ b/instruction/java-fundamentals/java-fundamentals.md
@@ -158,7 +158,7 @@ public class Person {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


leesjensen (commented):
Actually I still want to suggest that you clone the course instruction repository. It makes it a lot easier to get the dozens of files that you will need for the different phases. The instruction you are referencing is just to setup your repo, not to copy the starter code. 

Maybe this isn't clear enough that there is:

1. Your chess repo that you create
2. The instruction repo that has files you want to copy to stater code from

instruction/java-fundamentals/java-fundamentals.md
leesjensen, line 1: Thanks! I cherry picked this into main

PR #43, opened 2024-01-12
Fix Server.java: make compatible with Windows file paths and add clarification

prattnj:
## Add Windows Compatibility
Java's `Paths.get()` doesn't support colons in the filepath (for example, C:/), so `Server.java` was rewritten to support Windows users. `Spark.staticFiles` references the resources directory, so no instructions need modification.

## Add Clarification
Add a tip regarding opening a project in IntelliJ rather than creating a new one to avoid setup issues.

### DIFF
--- a/chess/4-database/starter-code/serverTests/PersistenceTests.java
+++ b/chess/4-database/starter-code/serverTests/PersistenceTests.java
@@ -58,6 +58,9 @@ public class PersistenceTests {
 
         // Restart the server to make sure it actually is persisted.
         stopServer();
+        try {
+            Thread.sleep(100);
+        } catch (InterruptedException ignored) {}
         startServer();
 
         //list games using the auth


chess/3-web-api/starter-code/server/Server.java
```diff
@@ -2,15 +2,12 @@
...
-import java.nio.file.Paths;
-
 public class Server {
 
     public int run(int desiredPort) {
         Spark.port(desiredPort);
 
-        var webDir = Paths.get(Server.class.getProtectionDomain().getCodeSource().getLocation().getPath(), "web");
-        Spark.externalStaticFileLocation(webDir.toString());
+        Spark.staticFiles.location("web");
```
leesjensen, line 10: The problem with this is that it doesn't work for both Maven and running in IntelliJ. You basically reverted it back to what I had before I made the change to support the autograder.

chess/0-chess-moves/getting-started.md
```diff
@@ -20,7 +20,7 @@ Take the following steps to set up your Chess project. Make sure you have alread
 
 ## Open With IntelliJ
 
-Open the project directory in IntelliJ start developing, running, and debugging your code using an IDE.
+Open the project directory in IntelliJ to start developing, running, and debugging your code using an IDE. Once you open IntelliJ, make sure you OPEN the project rather than creating a new project. The contents of the .zip file already contain IntelliJ configuration files and creating a new project rather than opening an existing one will cause various errors.
```
leesjensen, line 23: I like this. I'm going to push this directly into main.

chess/3-web-api/starter-code/server/Server.java
leesjensen: We need to figure out something that works on all the platforms.

prattnj: Are you sure this doesn't work in Maven? The autograder is in Maven and we haven't encountered any issues with this method.

leesjensen: It has been a month now since I worked on it, but when I ran it in intelliJ it set the execution directory to a different place than when I ran it with Maven. So I moved to it to being relative to the location of the class file. If you have tested all of those permutations and you are sure it works, then let's change it back to the simplified code. Just let me know that you have verified it.

leesjensen: @prattnj have you checked that your change works in all these environments:

1. Intellij execute run/debug for the client and server modules.
2. Maven running tests locally
3. Maven running tests in the autograder

If so then I would love to simplify this.

prattnj: It works for the autograder's own server, but I realize now that having multiple modules (the autograder itself doesn't) could be a variable that affects it. I try it but if it doesn't work, I believe I have another solution (using the `URI` class).

prattnj: I have tested and verified that `Spark.staticFiles.location("web");` works in the following situations:

- Running tests with Maven locally (Windows)
- Running tests in IntelliJ locally (Windows)
- Submitting a complete (or imcomplete) Phase 3 to the autograder, which uses Maven

leesjensen: Thanks for doing all this extra work!

prattnj: Michael just told me about the case in which you said it wouldn't work, which is when the client is involved. I'm not sure why anything would change, since the web socket tests are still in the server module but I will test that as well.

leesjensen:
@prattnj - Why did you close this? I don't think we have resolved the `web` directory problem yet? I'm still trying to try all the reproduction paths, but got stuck shaving a yak. I'm hoping to look at it more today.

prattnj:
That was inadvertent, a result of me opening a new PR while forgetting that this one was still active. I'm nearly done testing this with the client module, running into some other issues currently but hope to figure it out today.

leesjensen:
I have now verified that the following works for both IntelliJ and Maven on the mac. So I'm good with deploying your change.

```java
     Spark.staticFiles.location("web");
```

I cherry picked it into main. Thanks for your work on this, sorry I slowed things down.

PR #45, opened 2024-01-13
Corrections

frozenfrank:
In separate commits,
* Clarify wording
* Adjust the heading orientation. Previously, the header was centering itself in the column whereas the data was left aligned. I made them all left-aligned for consistency.

Previous Behavior:
image


leesjensen (approved):
Thanks!

PR #46, opened 2024-01-19
fix(chess4) bad dependency version number

frozenfrank:
## Overview
The old version of the phase 4 getting started told students to use a nonexistent version (`mysql:mysql-connector-java:8.0.2`) which results in an error shown below. I believe the document is simply missing an additional digit because `mysql:mysql-connector-java:8.0.20` is a valid version that will download. However, after checking with @pawlh, we discovered that the pass-off tests are actually using a more advanced version. Therefore, we're requested to update the instructions to show the same version currently used by the pass-off tests.

## Images
Screenshot of the Error:
image
_NOTE: In this screenshot, it shows `mysql` because I had installed it during testing._

Screenshot of success (when using the correct version number):
image

### DIFF
--- a/chess/4-database/getting-started.md
+++ b/chess/4-database/getting-started.md
@@ -30,7 +30,7 @@ This should result in the following additions to your project.
 
 Add the dependency for the MySQL driver and BCrypt. Associate them with your `server` module.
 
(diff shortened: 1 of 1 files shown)


leesjensen (approved):
Great catch! Thanks!

PR #47, opened 2024-01-19
Add time buffer to persistence test for slower machines

prattnj:
For slower machines, issues arise with stopping the server and immediately running the same instance. I added a line to let the process sleep for 100 milliseconds before restarting the server. 100 ms should be plenty of time, mine worked down to around 8.

### DIFF
--- a/chess/4-database/starter-code/serverTests/PersistenceTests.java
+++ b/chess/4-database/starter-code/serverTests/PersistenceTests.java
@@ -58,6 +58,9 @@ public class PersistenceTests {
 
         // Restart the server to make sure it actually is persisted.
         stopServer();
+        try {
+            Thread.sleep(100);
+        } catch (InterruptedException ignored) {}
         startServer();
 
         //list games using the auth


leesjensen (commented):
This is a great example of why global variables (and singletons) are usually unnecessary and not a good idea. Spark has a global Singleton (Spark.class).

```java
    private static class SingletonHolder {
        private static final Service INSTANCE = Service.ignite();

        private SingletonHolder() {
        }
    }
```

This makes it so that you can't start up two servers in the same code base. There is no real reason for this from a networking standpoint. If they had allowed it then we should be able to start up multiple server in a test and it would work fine. But because it is a global, and the actually do the stop asynchronously (which actually makes some sense) I can't start another one until the async work has completed.

However, just throwing sleeps in your code is a really bad smell that something else is wrong and you should go fix that. The sleep just is a poor bandaid and it will only sometimes work. I never see this problem, you solve it with 100ms, is there some other computer out there that needs 1000ms? You don't know.

Instead I propose this in the server code (Server.java):

```java
    public void stop() {
        Spark.stop();
        Spark.awaitStop();
    }
```

Now we know that when the stop method returns that we can safely start it up again.

leesjensen:
I added the `awaitStop` to Server.java starter code. Let me know if you don't think this will work. Since I can't repo this I am relying on you to test it for me.

prattnj:
I've tested it and it works! That's definitely a much cleaner solution than sleeping. I had tried `awaitStop()` previously but my problem was that I had put it in place of `stop()` rather than after it.

PR #48, opened 2024-01-19
enhance: clarify Object method overrides

ThanGerlek:
Clarify that only class fields are needed to generate methods, not necessarily a full implementation of the class.

### DIFF
--- a/chess/0-chess-moves/chess-moves.md
+++ b/chess/0-chess-moves/chess-moves.md
@@ -100,9 +100,13 @@ For this project, you are free to implement the classes described above in whate
 
 ## Object Overrides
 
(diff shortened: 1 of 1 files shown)


ThanGerlek:
There are a couple things I wasn't sure if I should include (especially since this is visible to students during the programming exam). I can remove them if you'd like:
- Specifying exactly which classes need overridden equals()
- Mentioning `deepEquals()`

frozenfrank:
Beautiful @ThanGerlek! That looks like it will be very helpful for the students.

leesjensen (approved):
Thanks! I reworked it to remove the exact directions of what they should do while still retaining the pointers that lead them on the right path.

PR #52, opened 2024-01-29
Added a test to check if the starting position is blank on a move

Truthwatcher64:
For the chess game tests there was a situation that was not originally tested. If someone was to pick a starting position for a move and that position was a blank square then weird things could potentially happen. Most likely the code would crash with a null pointer exception. This test makes sure that students have added a check to see if the starting position has an actual piece. If they find a blank square they should throw a invalid move exception.


leesjensen:
We can't make changes to the test for the current semester because the code is already live. So instead I have added this change for a branch that will be applied next semester.

https://github.com/softwareconstruction240/softwareconstruction/blob/2024winter/chess/1-chess-game/starter-code/passoffTests/chessTests/ChessGameTests.java

Thanks!

PR #58, opened 2024-02-01
Verify no dups array list

frozenfrank:
Requesting to merge into the `2024winter` branch to avoid confusion from multiple versions of the test cases this year.

This is a clone of softwareconstruction240/chess#4 which was recently merged. As a result of softwareconstruction240/chess#5, this code in a separate repository is now out of date. Here I present the PR to keep the changes in sync, but the real solution is to put all the relevant code into the same repository to avoid this kind of issue in the future.

PR #59, opened 2024-02-01
Validate that moves in ArrayList are all unique (Phase 1)

frozenfrank:
## Description
While uncommon an uncommon issue, without this adjustment, students may be accidentally adding the same move into their array list multiple times, and it would pass. If this is the case, then the test case should fail and point out their mistake so they can clean up their move logic.

## Notes
Requesting to merge into the `2024winter` branch to avoid confusion from multiple versions of the test cases this year.

This PR applies a change similar to the one already merged as softwareconstruction240/chess#4. Note that the change is required in two different places because the Phase 1 starter code is in a different location than the Phase 0 starter code (This is the topic of softwareconstruction240/chess#5).

### DIFF
--- a/chess/1-chess-game/starter-code/passoffTests/chessTests/ChessGameTests.java
+++ b/chess/1-chess-game/starter-code/passoffTests/chessTests/ChessGameTests.java
@@ -777,7 +777,9 @@ public class ChessGameTests {
 
 
     private void assertMoves(ChessGame game, Set<ChessMove> validMoves, ChessPosition position) {
-        var actualMoves = new HashSet<>(game.validMoves(position));
+        var generatedMoves = game.validMoves(position);
+        var actualMoves = new HashSet<>(generatedMoves);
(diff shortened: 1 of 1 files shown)


frozenfrank:
@leesjensen 

From [a comment](https://github.com/softwareconstruction240/softwareconstruction/pull/58#issuecomment-1921530866) in the cloned PR:
> It looks like you are trying to sync up the winter2024 branch for `softwareconstruction` to the lastest code through a pull request and then add some more changes to the instruction. 

I'm actually not trying to update the instruction, I'm adjusting the Phase 1 test cases. This is the topic of softwareconstruction240/chess#5. The test cases and other related starter code for Phase 1 and later are actually contained in the `softwareconstruction` repo, which is why I am forced to PR against this repo. Ideally, all of the starter code for phase 0 and phases 1+ would all be in the same repository, but that isn't currently the case.

leesjensen (approved):
Thanks!

PR #63, opened 2024-02-05
Remove unneeded starter-code links

ThanGerlek:
Removed links in phase 2 specs to starter-code.md and getting-started.md, which don't exist.


leesjensen (approved):
Thanks!

PR #64, opened 2024-02-07
fix(phase-1): expected and actual reversed for loadBoard() calls

ThanGerlek:
Every assertEquals() using loadBoard() had the expected and actual values flipped around. I checked every assertEquals() call in all 6 phases, and it looks like all the others are fine.


leesjensen (approved):
Doh! Thanks!

PR #65, opened 2024-02-07
Update chess-moves.md

jwsearle18:
The validMoves method fails when you have it honor who's team it is.

### DIFF
--- a/chess/0-chess-moves/chess-moves.md
+++ b/chess/0-chess-moves/chess-moves.md
@@ -69,7 +69,7 @@ public enum PieceType {
 
 #### Key Methods
 
-- **pieceMoves**: This method is similar to `ChessGame.validMoves`, except it does not honor whose turn it is or check if the king is being attacked. This method does account for enemy and friendly pieces blocking movement paths. The `pieceMoves` method will need to take into account the type of piece, and the location of other pieces on the board.
+- **pieceMoves**: This method is similar to `ChessGame.validMoves`, except it does not check if the king is being attacked. This method does account for enemy and friendly pieces blocking movement paths. The `pieceMoves` method will need to take into account the type of piece, and the location of other pieces on the board.
 
 ### ChessMove
 


leesjensen (commented):
Are you suggesting that pieceMoves does actually care whose turn it is? Meaning that it should return an empty collection if it is not the pieces color's turn?

leesjensen:
I think you are saying that there is a test that fails if your implementation of `ChessGame.validMoves` enforces whose turn it is. What make you think that this is true, and if so how does changing the definition of `ChessPiece.pieceMoves` relate to this?

jwsearle18:
Yes, sorry my wording wasn't the best.  What I am saying is that there are 3 tests that fail (at least in my implementation) if 'Chessgame.validMoves' enforces whose turn it is: Pinned King Causes Stalemate, Pieces Cannot Eliminate Check, and Check Forces Movement.  The reason I suggested this change is because as  'ChessPiece.pieceMoves' is defined, 'ChessGame.validMoves' is indirectly defined by differentiation, stating that 'ChessPiece.pieceMoves' doesn't enforce whose turn it is while 'ChessGame.validMoves' does.  In order to get a better understanding of how to implement 'ChessGame.validMoves', I read this and decided to enforce whose turn it is.  In doing so, these tests failed, so I figured 'ChessGame.validMoves' actually isn't supposed to enforce whose turn it is.  I decided to see if I could change this wording just in case this might cause issues for anyone in the future.  If you don't feel it's necessary, that's fine.  I just thought I might send it in just in case. 

I've uploaded screenshots of my failing and passing tests with the teamTurn check added and omitted.

Thanks for considering!

Screen Shot 2024-02-07 at 3 40 21 PM
Screen Shot 2024-02-07 at 3 40 53 PM

jwsearle18:
Screen Shot 2024-02-07 at 3 58 35 PM

Sorry this was coverd in the screenshots

leesjensen:
Thanks for the clarification and the proposed changes. 

My confusion is that your change is to remove the instruction for `ChessPiece.getMoves` that specifies that it should not consider whose turn it is. This would actually conflict with what the tests are doing.

I don't think there is any instruction for `ChessGame.getValidMoves` about if it is supposed to reference the player's turn, but I the tests definitely fail if you do enforce that.

Likewise for `ChessGame.MakeMove`. The tests for that will fail if you don't consider whose moves it is.

So I think we can do one of three things:

1. Change the tests and instruction so that `getValidMoves` enforces the turn.
1. Change the instruction so that it is clear that `getValidMoves` doesn't enforce the turn.
1. Do nothing since there is no conflict between the instruction and what the tests are requiring.

I think you only had a problem because you assumed `getValidMoves` should enforce the **turn**. The instruction change proposal you made doesn't seem to fix that, it only further confuses things because 1) the tests don't do what you assumed, and 2) the change would make the tests in conflict with the instruction.

Additionally,  the `ChessPiece.getMoves` function cannot use **turn** in its calculation because the board doesn't have a concept of **turn**. That is on the `ChessGame`.

What do you think?

jwsearle18:
I appreciate the follow-up, I'm sorry I haven't gotten back to you.  I see how my proposed change will further confuse things, as it is an edit to the instruction for 'ChessPiece.pieceMoves', and my proposed change might confuse people into adding a turn verification in this method.  I suppose no change is necessary since most students don't look back at the 'ChessPiece.pieceMoves' instruction for further clarification of the 'ChessGame.validMoves' method.  I was just confused because I looked for further clarification of the 'ChessGame.validMoves' method by looking at the instruction for the 'ChessPiece.pieceMoves' to see how they were different.  Based on what I read there, it looked like it was implied that one of the difference between these two methods was that 'ChessPiece.pieceMoves' doesn't verify team turn (which is true), while 'ChessGame.validMoves' does (which is false).  I can see now how it doesn't necessarily imply this difference.

Thanks so much!

leesjensen:
@softwareconstruction240/tas - Any thoughts about this?

Which of these would you prefer? Anyone with a strong opinion willing to do the work?

1. Change the tests and instruction so that getValidMoves enforces the turn.
2. Change the instruction so that it is clear that getValidMoves doesn't enforce the turn.
3. Do nothing since there is no conflict between the instruction and what the tests are requiring.

ThanGerlek:
I would suggest option 2, but I would change the instruction for ChessGame.makeMove as well, since it claims to provide a definition for "legal" that (misleadingly) only applies to itself. @jwsearle18, do you want to propose a change? Something like the following:

- validMoves: Takes as input a position on the chessboard and returns all moves the piece there can legally make. If there is no piece at that location, this method returns null. This method takes into account all applicable chess rules *except* that it does not take into account whose turn it is.

- makeMove: Receives a given move and executes it, provided it is a legal move. If the move is illegal, it throws an InvalidMoveException. This method takes into account all applicable chess rules: if you couldn't make a move in an actual chess game, makeMove should reject it.

- pieceMoves: Takes as input a chessboard and a position on the chessboard and returns all moves the piece there can make. If there is no piece at that location, this method returns null. This method does account for enemy and friendly pieces blocking movement paths. It will need to take into account the type of piece and the location of other pieces on the board. It does not take into account whose turn it is or whether the king will be left in check.

frozenfrank:
## Existing Question & Answer
@jwsearle18 It seems that your confusion has already been asked and answered in [this Slack Thread](https://cs240winter2024.slack.com/archives/C06C4187D2B/p1706305982673299?thread_ts=1706207420.516449&cid=C06C4187D2B).

I'll excerpt the conversation here for clarity and preservation:
> **Student:** According to the spec, validMoves is supposed to return only legal moves, and legal moves are ones made when it's the correct team's turn. But in order for the validMoves tests to pass, validMoves must not check for the current team turn because the tests don't actually set the team turn, so it's null. The check for current turn needs to be done in makeMove instead.
...

> **Student:** validMoves is a member of ChessGame though, not ChessPiece or ChessBoard. So wouldn't it still be good practice for it to check the turn (if it worked, of course)?

> **TA:** No, it would not be good.
First of all, it would be a duplication of responsibility.
Second of all, it would prevent you from legitimately using the method to identify valid moves for a piece in alternative contexts. For example, later in phase 5 you’ll be responsible for highlighting all the moves a piece can make on the board. A user may want to explore what moves are available even if it isn’t their turn (they may be waiting for the other player to make a move). Therefore, you would be unnecessarily preventing a user from a legitimate behavior by validating the turn inside validMoves().

## Chess Organization
The tests are treating the functions just the way they should. Specifically:
- **Only** `ChessGame.makeMove()` should enforce the current turn of the game
- `ChessBoard.validMoves()` should return all the moves a player could make _assuming it were their turn_, taking in to account the possibility of exposing the king to check
- `ChessPiece.pieceMoves()` should return all the moves a piece could make just by mechanically following the rules of it's movement (but without consideration of higher level game rules)

This distribution of responsibility does correctly implement a chess engine, and it also minimizes the duplication of responsibility and the organization of good code.

## Therefore
Regarding the proposals of @leesjensen:
> ~1. Change the tests and instruction so that getValidMoves enforces the turn.~
> 2. Change the instruction so that it is clear that getValidMoves doesn't enforce the turn.
> 3. Do nothing since there is no conflict between the instruction and what the tests are requiring.

1. Not an option. The tests are correct and the code organization is good.
2. Feasible, but this PR doesn't address the issue.
3. Also feasible, but future students may continue to be confused about this point (we did have to answer similar questions about this multiple times in the Slack channel).

This PR should be closed.

frozenfrank:
Hey @jwsearle18! We are interested in helping you contribute to the class and make it better, but we'll need you to participate with us. We're closing this PR for now. If you still feel like this is an important issue, please respond back in the comments or open up a new PR to address the issues.

Issue #70, opened 2024-02-12
Phase 2--UML template (and example image) don't match specs


ThanGerlek:
In a couple of places, the provided template UML diagram (and the example image derived from it) don't match the specs.

- The `register` endpoint returns the authToken on success, but not the username
- The `joinGame` request has a `clientColor` variable but the tests need it to be named `playerColor`

ThanGerlek:
Closed by PR #71

Issue #67, opened 2024-02-12
Phase 2 Starter Diagram: Generalize RegisterService to Service.


frozenfrank:
## Overview
Change the value of the middle layer in the example and in the starter file to just show "Service" instead of "Register Service."

## Discussion
The main point of the diagram is to think through and understand the relationship between the different layers of your server code at a high level. It’s sufficient to think about it merely as “Server,” “Services,” and “DataAccessObjects” where there is one Server, multiple “Services” (ex: RegistrationService and GameService) and multiple “DataAccessObjects” (ex: UserDAO, GameDAO, AuthDAO). However, the over-specifity of the example interferes with this generalization. Additionally, the starter code uses this same title, but the whole column is reused for all of the methods.

leesjensen:
Done.

PR #71, opened 2024-02-12
Update UML template diagram (#70, #67)

ThanGerlek:
Update template to match specs (fixing #70) and generalize RegisterService to Service (implementing #67). This includes updating example.uml and register-sequence-diagram.png (but not the .gif).


leesjensen (approved):
Thanks!

PR #72, opened 2024-02-14
Update web-api.md

Truthwatcher64:
Changed the Code snippet to match the code that is in the starter code.

The Markdown has an older version of the code. The new snippet will work and is the one that is from the starter code.


leesjensen (approved):
Thanks!

PR #74, opened 2024-02-21
Fixed code in Petshop server that reads the db.properties file to work on Windows.

rodham:
The Petshop code that reads db.properties was not working on Windows due to the "C:" drive name in the value returned by DatabaseManager.class.getProtectionDomain().getCodeSource().getLocation().getPath(). I reimplemented this code in a way that will work on all OSes.


leesjensen:
Thanks

leesjensen:
We probably need to fix this on the chess starter code also

leesjensen:
Thanks

PR #77, opened 2024-02-28
Clearly mark Phase 3 Rubric requirements that result in a rejected submission

### LINKED ISSUE #76: Phase-3 Rubric doesn't clearly distinguish grades from prerequisites
The phase-3 rubric isn't explicit that not having enough unit tests will result in a rejected submission. It is currently included in the `Unit Tests` section, which is listed as being worth 25 points, when in truth that doesn't apply since that's an absolute requirement. Having no unit tests is technically -180 points but the current rubric implies that it's only -25. The same problem applies (to a lesser extent) to the requirement that the webpage loads, which is listed as worth 10 points, but in theory should result in a complete rejection.

Another resolution would be to simply change the auto-grader to not reject submissions without unit tests, so it is actually worth -25 points.

ThanGerlek:
Split the `Unit Tests` section, worth 25 pts, into `Unit Tests` (marked as a Prerequisite) and `Unit Test Quality` (25 pts).

Also marked `Web Page Loads` as a Prerequisite and moved its 10 points into `Web API Works`.

Resolves #76


chess/3-web-api/web-api.md
```diff
@@ -417,8 +417,9 @@ After your code has successfully been auto-graded, a TA will review the code in
...
 | :------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -----------: |
 | GitHub History | At least 10 GitHub commits evenly spread over the assignment period that demonstrate proof of work                                                                                               | Prerequisite |
-| Web API Works  | All pass off test cases in `StandardAPITests.java` succeed                                                                                                                                       |          115 |
-| Web Page Loads | Test web page properly loads in browser (including all files linked to by the test page: favicon.ico, index.css, index.js)                                                                       |           10 |
+| Web Page Loads | Test web page properly loads in browser (including all files linked to by the test page: favicon.ico, index.css, index.js)                                                                       | Prerequisite |
```
leesjensen, line 420: Do we really need this? I assume the point of removing the 10 point award specifically for the web page is due to the new test that makes sure it loads, and therefore is included in the 125 points for the tests. We don't call out any other functionality from the tests as a specific prerequisite. Why this one?

ThanGerlek:
Will not fix

PR #78, opened 2024-02-28
fix: update description of data model class locations

ThanGerlek:
Clarify that Data Model classes don't need to be in the chess package (and fix a couple typos)


leesjensen (approved):
Thanks!

PR #85, opened 2024-03-05
Add automatic scrolling for API test page

19mdavenport:
This adds automatic scrolling to the execution area when a endpoint button is clicked to the API test page. Many students are initially confused how to use the test page and I believe this would help them understand it more quickly and use it more efficiently.


leesjensen (approved):
Thanks!

PR #93, opened 2024-03-11
Clarify that `db.properties` must be used

pawlh:
Currently the Phase 4 specs do not explicitly say that `db.properties` must be used, it is only implied.

This change adds a clarification that when creating the database, the values must come from `db.properties`


leesjensen (approved):
Looks great!

PR #95, opened 2024-03-14
Use correct reset escape codes

19mdavenport:
This fixes the EscapeSequences class to have what I believe is the correct codes for resetting colors. The existing RESET_BG_COLOR and RESET_TEXT_COLOR just set both to be black


pawlh:
Can confirm that this works on the IntelliJ terminal 🥳

leesjensen (approved):
thanks

PR #98, opened 2024-03-15
90 enhance exception instruction

### LINKED ISSUE #90: Enhance instruction about using Exceptions
Details to add/improve upon:
- Using multiple Exception types/subclassing
- Where to catch and where to use `throws`
- Warn students not to rely on IntelliJ's auto-generated try/catch blocks (always re-write the catch)

ThanGerlek:
Resolves #90. I intentionally added more than was necessary but I figured it's easier to remove than add. It also might be too opinionated in places; it definitely encourages using exceptions much more freely than before. I suggest multiple people review it before merging


leesjensen:
@ThanGerlek - Thanks for proposing all of these. I'm still thinking about how to properly implement this and enhance the instruction.

PR #99, opened 2024-03-15
feat: add warning about case-sensitivity of table names

### LINKED ISSUE #89: Add note to Phase 4 spec about SQL table name sensitivity
Different operating systems do or don't care about case in SQL database names and table names (and maybe column names...?). We should add a warning about this

ThanGerlek:
Resolves #89


leesjensen (approved):
thanks!

PR #101, opened 2024-03-22
enhance(phase-5): require that chess boards "look like a chess board"

ThanGerlek:
Require drawn boards to "look like a chess board" according to official chess rules (have alternating colors, white-on-the-right, etc.)


leesjensen (approved):
Looks great!

PR #108, opened 2024-04-05
Add check and checkmate notifications to "Game Completion" rubric item

ThanGerlek:


leesjensen (commented):
I'm not completely sure what the tests require for this. But it seems like this needs to be clarified in the **Root Client sends MAKE_MOVE** section since that this where the notification is generated. Perhaps a line that says:

> If check or checkmate results from the move then the server sends a `Notification` message to all **clients** in that game informing them of check or checkmate.

This would clarify the information provided in the notifications section.

That then leaves the question of where to put this in the rubric. Should it be under the **Player Move Piece** item or the **Game completion** item.

jerodw:
I added the notification to the Player Move Piece item instead, so I'm rejecting this PR.

PR #113, opened 2024-04-30
Update Everything to New Phase 6 Design

prattnj:
# In This PR

Throughout the project, the specs/starter code/instructions/petshop code have been updated to match the following rules:
- In phase 6, the `JOIN_PLAYER` and `JOIN_OBSERVER` `UserGameCommand`s have been merged into one called `CONNECT`. It only requires a `gameID` in addition to the other fields included in `UserGameCommand`.
- In phase 3, the Join HTTP method (`PUT /game`) no longer accepts null values for the color. A null color should return `400` (though the tests have not been changed yet). In other words, there is no such thing as "observing" a game via HTTP.
- In phase 5, the options for joining as a player and as an observer are still in the post-login menu, but the language has been updated and joining as an observer has no functionality until phase 6.
- The endpoint that the WebSocket connects with is now `/ws` rather than `/connect` to avoid confusion in phase 6.

In addition to these rule updates, wording in the specs has been updated so that the word "join" can ONLY refer to the HTTP join.


prattnj:
> @prattnj In your first commit you're removing Dr. Rodham's websocket diagram because it's outdated. What specifically is outdated? I think he's going to want to update it.

In his diagram, he references JoinPlayer and JoinObserver, which we replaced with one command called Connect. I plan on updating the diagram, but for the sake of quickly updating things for this term, I temporarily removed it in the mean time.

jerodw:
I'd rather not delete his diagram, in case we forget to replace it. It seems like a simple change. You can either ask him to update it and send you the update diagram to include in another commit or maybe he can provide the original so you can update it.

prattnj:
> I'd rather not delete his diagram, in case we forget to replace it. It seems like a simple change. You can either ask him to update it and send you the update diagram to include in another commit or maybe he can provide the original so you can update it.

I can just remake it, there's plenty of time since the queue isn't too busy.

jerodw:
He may want to have the original. I'll ask him if he wants to update it or if he wants you to remake it.

prattnj:
> He may want to have the original. I'll ask him if he wants to update it or if he wants you to remake it.

I just updated it, it wasn't as difficult as I thought because apparently the software I use for this kind of thing can edit PDFs.

prattnj:
Since Dr. Rodham moved the diagram to a better location (the Phase 6 specs directory), I removed it from its location here. It isn't actually deleted, just moved. It should be ready for merging now.

PR #114, opened 2024-05-02
Removed Starter Code Dead Links (Phase 1-6)

Maillman:
The links to the starter code on the various .md files from Phase 1-6 don't lead anywhere. I just replaced the non-functional links with text saying the starter code is in the Phase 0 zip.


leesjensen:
Thanks for noticing this. I went and removed the references entirely. This seemed cleaner since there is already text referring them to their chess project starter code directory rather than pointing them back to the original chess.zip.

PR #116, opened 2024-05-14
Update to Phase 2

Truthwatcher64:
Added the use of the handler to the uml diagram.


jerodw (commented):
The changes look good but we need to make a couple of changes to the UML diagram. The line from handler to service is a method call that has a Register request as a paramter, so it should be labeled register(RegisterRequest). The return line coming back from Service to Handler is a return, not a method call, so it should be labeled RegisterResult (without the parantheses).

PR #117, opened 2024-05-16
Clarify Specifications

19mdavenport:
This clarifies:

- the role of isInStalemate
- the success responses for the clear, logout, and join endpoints
- the error messages for 500 errors
- the type of the `game` field in `GameData`

### DIFF
--- a/chess/3-web-api/web-api.md
+++ b/chess/3-web-api/web-api.md
@@ -127,13 +127,13 @@ As part of this phase, you need to create [record](../../instruction/records/rec
 
 **GameData**
 
-| Field         | Type                       |
-| ------------- | -------------------------- |
-| gameID        | int                        |
-| whiteUsername | String                     |
-| blackUsername | String                     |
-| gameName      | String                     |
-| game          | `ChessGame` implementation |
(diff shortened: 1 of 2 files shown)


chess/3-web-api/web-api.md
jerodw, line 1: I'm not sure about this commit. Is there a reason we need to require an empty JSON response in these cases instead of just no body?

jerodw, line 1: Why add code tag just to 500 responses? Isn't it just as relevant for the others?

19mdavenport: What would no body look like in code? Spark has issues if you return null from a Route. Returning an empty String works with the current tests

19mdavenport: I'm probably showing my relative inexperience with markdown format. Doesn't the code tag do roughly the same thing as backticks? I wanted to be able to italicize something in the code block, which is one difference between the two.

jerodw: Or I'm showing my inexperience with it. This is fine. I'll merge the PR

PR #119, opened 2024-05-23
Added a section to the getting started for phase 3 

Truthwatcher64:
I added a section that explains the webpage the server will use for testing. It's at the end of the getting started page.


leesjensen:
It looks like we lost the instruction to set up the server main function to call server.run. That probably needs to be added independent of this suggestion. After that the web page should just work. There are a couple references to the web page in the instruction, but it might be nice to have a final one here.

There is some discussion about if the web page is really necessary. Perhaps it should just be replaced with an emphasis on using curl instead.

PR #120, opened 2024-05-28
Clarify Phase 3 and 6 Serialization and Deserialization

Maillman:
Students generally have problems when serializing and deserializing Gson and Json objects, not understanding that the field names they implement must be exact (including capitalization). I just added a clarification about this in phases 3 and 6 where students have to serialize and deserialize the Json.

### DIFF
--- a/chess/3-web-api/web-api.md
+++ b/chess/3-web-api/web-api.md
@@ -13,6 +13,8 @@ In this phase, you will create your Chess server and implement seven HTTP endpoi
 
 An endpoint is a URL that your server exposes so that clients can make Hypertext Transfer Protocol (HTTP) requests to your server. Often the server requires some data when a client calls an endpoint. For an HTTP request this data can be stored in HTTP Headers, in the URL, and/or in the request body. The Server then sends back data to the client, including a value in the HTTP Response Code (indicating if the command was completed successfully), and any needed information in the HTTP Response Body. For your server, you will use JSON strings to encode the objects we include in the Request and Response bodies.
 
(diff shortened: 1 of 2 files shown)


leesjensen:
I can see that a few students each semester might run into this problem, but I worry that this might be a bit intrusive to the flow of the instruction. I worry that if we add a paragraph for every little thing a student could mess up, the instruction would quickly become unreadable.

This seems like this might be a good learning experience for those who get it wrong and it shouldn't be too hard to debug.

How many students have you interacted with that had this problem, or is this just based on a single student interaction?

Maillman:
I had this issue when I was taking the class myself, and it was quite of a headache trying to resolve. I have talked with two students taking about 20-30 minutes each who have had this problem before we managed to nail down the problem, I am unsure about the other TAs though. I do know that last semester this was a big issue especially during phase 6 as the results were very vague when trying to run the WebSocketTests. Really, I just pulled the info from 🖥️ [Slides: Server Implementation Tips](https://docs.google.com/presentation/d/1hORd88ej8W-nqHgEpYU2GmPcrSrHew1V/edit?usp=drive_link&ouid=110961336761942794636&rtpof=true&sd=true) JSON Tips from the web-api.md, and added just a little bit of clarification on why not doing this could lead to a bug.

Maillman:
I could also ask the other TAs and Dr. Wilkerson to get their opinion on this during the next meeting as well, if you want.

Maillman:
Never mind, it already does talk about having the fields needing to match exactly in the specs, just under the serialization portion. There is no need to specify it twice.

PR #122, opened 2024-05-31
Update chess zip with new database tests

19mdavenport:


19mdavenport:
Waiting until spring for this change. We will need other changes for spring so this PR will become outdated.

PR #123, opened 2024-06-05
Specify Code Quality for Phases 3-6

Maillman:
Now that code quality is required in all phases from 3-6, I just specified so in web-api.md and removed the duplicate (Code Quality) section in gameplay.md. It can probably either be this way or have a (Code Quality) section in each of the phases' markdown.


Maillman:
Also, there needed to be a space character in pregame.md after a period.

PR #134, opened 2024-08-10
Update Phase 6 Rubric

19mdavenport:
This changes the rubric wording for highlighting moves to not care about current turn and for the passoff test cases to reflect that there are no longer 20 test cases (currently there are 17, I had to remove a few when we condensed JOIN_PLAYER and JOIN_OBSERVER to CONNECT. I could try to add more if you'd like to bring that back up to 20 for easy math).


leesjensen (approved):
Great!

PR #140, opened 2024-09-21
update(concurrency.md) Fully Update Concurrency Example

### LINKED ISSUE #110: Partially updated example in Concurrency module
Referring to [this file](https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/concurrency/concurrency.md?plain=1#L343).
It looks like in [this commit](https://github.com/softwareconstruction240/softwareconstruction/commit/50ab8d37648f1e0a52d7d917cd00f564cb1873bf) Prof. Jenkins changed the multithreaded HTTP server example from toggling a boolean `value` variable to adding and subtracting 1 from an integer `sum` variable. However, the following explanation for the example was never updated so it still refers to toggling the old `value` variable and doesn't make much sense.

frozenfrank:
Resolves #110.

Before submitting this PR, I recreated the sample to personally experience the effects of the code. This helped to provided a more accurate and clear description of the actual problem in the text. In testing, I extended the code with a few simple endpoints to fully support the operations needed.

## Full Testing Code
```java
public class MultithreadedServerExample {
    static int sum = 0;

    Spark.get("/add/:value", (req, res) -> {
        sum += Integer.parseInt(req.params(":value"));
        return " " + sum;
    });
    Spark.get("/set/:value", (req, res) -> {
        sum = Integer.parseInt(req.params(":value"));
        return " " + sum;
    });
    Spark.get("/get", (req, res) -> " " + sum);
}
```

```bash
curl localhost:8080/set/0
for i in {1..100}; do curl localhost:8080/add/1; print ""; done &
for i in {1..100}; do curl localhost:8080/add/-1; print ""; done &
wait
curl localhost:8080/get; print"";
```

### DIFF
--- a/instruction/concurrency/concurrency.md
+++ b/instruction/concurrency/concurrency.md
@@ -360,15 +360,15 @@ public class MultithreadedServerExample {
 }
 ```
 
-In this example, the `value` variable is read and modified in the endpoint handler. If you execute the following commands from your command console it will repeatedly try to toggle the value. After each call to toggle the value should return to true.
+In this example, the `sum` variable is read and modified in the endpoint handler. If you execute the the following commands from your command console, it will rapidly send 100 add requests and 100 subtract requests to the server. The requests may not fire at exactly the same rate so the intermediate values may fluctuate. By the end of the operation, once all the operations have completed, we would expect for the sum to be back at 0.
 
 ```sh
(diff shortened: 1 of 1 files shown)


leesjensen (approved):
Looks great. Thanks for the clearer explanation and the reduced execution from the `while` loop (which required you to terminate the execution). If 100 is enough to show the problem that is better.

PR #142, opened 2024-09-24
Update programming-exam-review.md

frozenfrank:
Indicate that students must have an Examity supported device.

### DIFF
--- a/instruction/programming-exam-review/programming-exam-review.md
+++ b/instruction/programming-exam-review/programming-exam-review.md
@@ -13,6 +13,7 @@ You will use Examity to record both video and audio while taking the exam.
 ## Examity Exam Requirements
 You must meet the following requirements for taking Examity exams:
 
+- Have a [supported device](https://www.examity.com/test-taker-systems-requirements/). Note that Examity does not support Linux devices.
 - Have a relatively new version of the Chrome web browser (other browsers will not work)
 - Have only one monitor attached to your computer (if you have additional monitors, you will need to unplug them)
 - Have a working video camera and microphone


leesjensen:
@frozenfrank , @jerodw  - This feels like the wrong direction. If we want top notch engineers graduating from BYU then Linux is something we should be supporting not discouraging. Could we instead include instructions about how to get Examity to work with Linux. Spoof the user-agent and such. Alternatively say that they can come into the TA lab and take the test there? I think we have room for this in the Lab.

frozenfrank:
@leesjensen The main motivation for this change is an incident yesterday where a student was trying to spoof the user agent, but the screen sharing still wasn't working properly and issues arose. 

## Student Examples
Spoofing the user agent did work for [this student](https://cs240-fall2024.slack.com/archives/C07KV0UFEP2/p1727200843911459).

See the [Slack thread here](https://cs240-fall2024.slack.com/archives/C07KV0UFEP2/p1727190850418009) for the student with issues:
> **Student:** I need urgent help right now. I can't get my screen to be broadcast. When I tried the practice exam yesterday, it worked just fine. But today, my DE is failing to broadcast the screen when prompted. I'm using debian/KDE.
> For some reason, I'm able to broadcast another tab of Chrome, but not anything on my desktop
> I just got off a meeting with a TA on the help queue, but he didn't know what to do, and suggested I post here. Is anyone seeing this? I'm running out of time to take this test
> I get the message, 'Failed to create Pipewire Context,' but no error logs are generated
> I've tried updating everything and restarting my computer
> **Fellow Students:** Attempt to help, but have no meaningful information
> **Student:** I hope so, because I'm now out of time. Or rather, it would be very unlikely that I could finish the test before my next obligation if I were to start right now.

## What's the Bigger Issue?
In terms of supporting Linux users, the bigger roadblock here is Examity's support and not our instructions referencing Examity. Changing our instructions here won't entirely resolve issues with students using unsupported operating systems. 

This is really an issue about inclusion (DEI) and belonging than the instructions here. The same student who had issues with Linux also is not able to use Intelli-J and faced issues when trying to build the project outside of that particular IDE. The best forward will make the programming exam more platform independent and also make the chess project more IDE independent.

## Alternative Directions
Alternative directions could include:
* Switching away from Examity towards a platform that supports Linux
* Allowing Linux users to skip the Examity proctoring
    * We could offer in-person TA proctoring as an alternative
* Suggest borrowing a mac/windows machine for the exam
    * The library or other campus resource may offer devices
    * Students have borrowed devices from roommates/friends before
* Some other solution.

## Moving Forward
@leesjensen Would you like me to move the information into a new issue so we can continue discussing it in a TA meeting as we attempt to get better?

jerodw:
Although I agree that it would be nice to support Linux users, this sentence is not moving away from supporting it. It's just stating what we've always said in class.

BYU Online has a contract with Examity and will not support or pay for any other exam monitoring solution in BYU Online exams, so I don't see moving away from Examity to be an option. We tried before when we wanted to stay with Proctorio.

I don't think we should be using much TA time trying to support what Examity doesn't support. Maybe we can suggest three options, in this order of preference:

1) Use a Mac or Windows machine if possible (recommended)
2) Provide instructions on spoofing the user agent and tell them this has worked for some but not all Linux users, and TAs are not able to provide support for it
3) Talk with your instructor (who can make arrangements for TA proctoring of the exam. In that case students could take the practice exam while being monitored by a TA)

If we do this I wouldn't want to provide details about what the instructor might do for option 3, just tell them to talk with their instructor, so we don't have a flood of students asking for TA proctoring.

PR #144, opened 2024-09-28
update(chess-moves.md) Clarify validMoves requirements

frozenfrank:
Clarify requirements for `ChessGame.validMoves()` and `ChessGame.makeMove()` inside the Phase 1 specs. This provides students the details of which conditions should be verified in each method to represent the rules of chess.

Students have been able to figure this out by reasoning through the issue, but students less familiar with chess typically face a high burden here. Additionally, students are likely to misapply the verification of the current team's turn which— less intuitively— **should not** be verified in `ChessGame.validMoves()` because the `highlightMoves()` method needs to rely on it. However, this information isn't presented to students until much later in the project.

This change was peer reviewed by @Truthwatcher64

### DIFF
--- a/chess/0-chess-moves/chess-moves.md
+++ b/chess/0-chess-moves/chess-moves.md
@@ -82,7 +82,7 @@ classDiagram
 This class serves as the top-level management of the chess game. It is responsible for executing moves as well as recording the game status.
 
 > [!IMPORTANT]
-> The ChessGame class is presented now, but will not be used until Phase 1.
+> Although the `ChessGame` class is presented now, it will not be used until Phase 1.\
 > ChessGame related tests will be [added to the working directory](../1-chess-game/getting-started.md#chess-game---getting-started) later.
 
 
--- a/chess/0-chess-moves/chess-moves.md
+++ b/chess/0-chess-moves/chess-moves.md
@@ -81,15 +81,10 @@ classDiagram
 
 This class serves as the top-level management of the chess game. It is responsible for executing moves as well as recording the game status.
 
-You will not need to implement any code in this class for this phase.
+> [!IMPORTANT]
+> The ChessGame class is presented now, but will not be used until Phase 1.
+> ChessGame related tests will be [added to the working directory](../1-chess-game/getting-started.md#chess-game---getting-started) later.
 
-#### Key Methods
-
-- **validMoves**: Takes as input a position on the chessboard and returns all moves the piece there can legally make. If there is no piece at that location, this method returns `null`.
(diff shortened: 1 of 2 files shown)


leesjensen:
My thought here is to just delete it rather than simplify it as suggested by #145.

frozenfrank:
@leesjensen Note that this PR is independent of #145 and focused on a different issue. This PR is directly clarifying the documentation for the required methods in the Phase 1 spec. This PR does not address the duplication or any of the larger organizations issues brought up in the separate issue. Notably, this PR is not changing the information in both of it's duplicated locations, only the one location that matters.

Prior to this change, there is a lack of clarity in the specs about which conditions should be verified in each method. The docs for some methods are clear about what conditions need to be checked (while still requiring the students to figure out how to implement the verifications), but the subject of this PR is not clearly stated.

It's possible that the information was omitted by assumption that students would be able to figure it out. Certainly, students familiar with the chess are able to reason and be wrong and figure out what is required by the tests. However, other students less familiar with the rules of chess are faced with a high barrier to figuring out what they're supposed to do.

leesjensen:
I guess I feel that the two are related. What I would like to do is:

1. Remove ChessGame from the phase 0 discussion
2. Enhance Phase 0 to say that the only rules that are implement here is how a piece moves. That is independent of the rules of a game such as check, stalemate, or checkmate.
3. Enhance Phase 1 with the clarification that you were proposing for phase 0 to the existing ChessGame information. That now you need to implement the rules of the game that include check, stalemate, and checkmate. That means you need to remove rules returned from the chesspiece that violate the game rules.

When I explained it in class the students seem to get the difference.

As a side thought. I want to keep away from is making any changes that require code changes either to the starter code, or anybody's implementation.

frozenfrank:
> This PR is directly clarifying the documentation for the required methods in the Phase 1 spec.

> Enhance Phase 1 with the clarification that you were proposing for phase 0 to the existing ChessGame information. 

@leesjensen I had _intended_ to propose these changes to the Phase 1 spec, but by mistake, make them to the Phase 0 specs instead. That has now been corrected.

I'll look more into the other adjustments requested tomorrow.

ghprime (approved):
LGTM! Thanks for making the instructions for `ChessGame` more explicit

frozenfrank:
@leesjensen Done.

The PR now reflects all the adjustments you requested. Note the following details:
* I removed the details of `ChessGame` from the Phase 0 spec, but it is still mentioned in the diagram and in the class list. This was done to describe its role in the broader class context; since the file is still in the starting working directory, it should also be mentioned in the specs.
* Based on feedback from other TA's, we also included a line about the required initial state of the `ChessGame` class. Students would figure it out by the clear test cases, but we felt it would be appropriate to comment on this here.

Are there any other adjustments you would like to see to these areas of the spec?

leesjensen (approved):
Love it. Thanks!

PR #147, opened 2024-10-15
Update web-api.md specifically the service class examples.

AldenKim:
Students were confused with the example of the service class passing in data model classes and returning data model classes. Per Me and Alex's recommendation, we would like to update this section to use the proper Request and Result objects.

### DIFF
--- a/chess/3-web-api/web-api.md
+++ b/chess/3-web-api/web-api.md
@@ -207,7 +207,7 @@ public class UserService {
 }
 ```
 
-Each service method receives a **Request object** containing all the information it needs to do its work. After performing its purpose, it returns a corresponding **Result object** containing the output of the method. To do their work, service classes need to make heavy use of the Model classes and Data Access classes described above.
+Each service method receives a Request object containing all the information it needs to do its work. After performing its purpose, it returns a corresponding Result object containing the output of the method. These request and result objects would contain fields pertaining to each of the endpoints above. (Note: request and result classes do not need to be specifically created to what is shown in the specifications, it is up to your design on how you want to implement these methods). To do their work, service classes need to make heavy use of the Model classes and Data Access classes described above.
 
 ⚠ You must place your service classes in a folder named `server/src/main/java/service`.
 
--- a/chess/3-web-api/web-api.md
+++ b/chess/3-web-api/web-api.md
@@ -201,13 +201,13 @@ A simple implementation of this is to have a separate Service class for each gro
 
 ```java
 public class UserService {
-	public AuthData register(UserData user) {}
-	public AuthData login(UserData user) {}
-	public void logout(AuthData auth) {}
+	public RegisterResult register(RegisterRequest registerRequest) {}
+	public LoginResult login(LoginRequest loginRequest) {}
+	public void logout(LogoutRequest logoutRequest) {}
 }
 ```
 
-Each service method receives a Request object containing all the information it needs to do its work. After performing its purpose, it returns a corresponding Result object containing the output of the method. To do their work, service classes need to make heavy use of the Model classes and Data Access classes described above.
+Each service method receives a **Request object** containing all the information it needs to do its work. After performing its purpose, it returns a corresponding **Result object** containing the output of the method. To do their work, service classes need to make heavy use of the Model classes and Data Access classes described above.
 
 ⚠ You must place your service classes in a folder named `server/src/main/java/service`.
 


leesjensen:
Why are they confused? Is it because you have the model in your mind that the services should all have "request" and "response" objects and that doesn't match what that example gives? Perhaps instead of replacing one way of doing it with a different one we should give a second example of a possible way to implement the code.

19mdavenport:
I think the removed example is valuable, what would you think about having both as examples? That would carry the downside of longer specs

AldenKim:
I can add both examples. Should we keep the description below the example the same as the original one, or should we keep those changes as well?

leesjensen:
I reread the entire Phase 3 instruction again today and realized that there is a discussion contrasts using specialized classes with using the model objects directly below the section discussed in this PR. I don't think we want to have that discussion twice.

image

frozenfrank:
> ... that there is already a discussion that contrasts using specialized classes with using the model objects... I don't think we want to have that discussion twice.

This observation leaves us with (_only_) two options for the example provided earlier in the document:
1. Present the code example with `Request` and `Response` objects, OR
2. Present the code example with `UserData` and `AuthData` objects.
3. ~Present the two code examples, one with each potential path~ _(this will be addressed later in the document)_

frozenfrank:
## Acknowledgments
We acknowledge that Prof. @leesjensen is a big proponent of making students think like software developers and avoid giving them the "only one right and acceptable" answer. In reality, there are multiple valid approaches, along with some common industrially accepted cheats to ideal organization, that all achieve the same result in the end. Students should be equipped to:
1. Be aware of the diversity of approaches
2. Choose for themselves how they would like to navigate the valid decision space

With that said, we already established that this particular point in the document is not the right place to have this kind of discussion. We acknowledge the the section below titled [Result and Request Classes](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/3-web-api/web-api.md#request-and-result-classes) _may be_ a good location for such an expanded written discussion. May I suggest, however, that this kind of perspective-shifting content is best delivered in class by a dynamic professor who can adapt to the diverse backgrounds and aptitudes of the students.

## Recommendation
In this particular place, it seems most consistent with the surrounding text, and to be consistent with the [starter Phase 2 diagram](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/2-server-design/server-design.md#register-example), that we **proceed with the changed already proposed** and refer to the parameters in this instances by `Register` and `Request` classes.

Students are obviously free to listen to perspectives presented in class, and to read the discussion in the "Request and Result Classes" section, and go a different route if they so desire.

## Works Cited
As was previously stated by [a wise contributor](https://github.com/softwareconstruction240/chess/issues/27#issuecomment-2383660639):
> Providing multiple instruction paths usually leads to confusion. 

😉

chess/3-web-api/web-api.md
```diff
@@ -201,13 +201,13 @@ A simple implementation of this is to have a separate Service class for each gro
...
-	public void logout(AuthData auth) {}
+	public RegisterResult register(RegisterRequest registerRequest) {}
+	public LoginResult login(LoginRequest loginRequest) {}
+	public void logout(LogoutRequest logoutRequest) {}
 }
 ```
 
-Each service method receives a Request object containing all the information it needs to do its work. After performing its purpose, it returns a corresponding Result object containing the output of the method. To do their work, service classes need to make heavy use of the Model classes and Data Access classes described above.
+Each service method receives a Request object containing all the information it needs to do its work. After performing its purpose, it returns a corresponding Result object containing the output of the method. These request and result objects would contain fields pertaining to each of the endpoints above. (Note: request and result classes do not need to be specifically created to what is shown in the specifications, it is up to your design on how you want to implement these methods). To do their work, service classes need to make heavy use of the Model classes and Data Access classes described above.
```
frozenfrank, line 210: These additions may be better replaced with a link to the [Request and Result Classes](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/3-web-api/web-api.md#request-and-result-classes) section which properly discusses this topic at length.

It may be appropriate to adjust the language in the final sentence of this line as well to be consistent with the direction of the PR. It makes reference to "Model classes" and "Data Access" classes, and there may be other ways to represent the relationship of all the related words.

frozenfrank (changes requested):
I used the language "proceed with the changed already proposed" in my earlier comments because I agree that this is a good direction, but I believe there may be better words to use to more completely capture the spirit of the direction being presented.

My thoughts on the issues are shared in the comments.

PR #152, opened 2024-10-29
Discrepancy in Phase 4 specs (super small)

AldenKim:
A student noticed that in the example for hashing a password, the function passes in a String password, but inside the actual function we use clearTextPassword, just wanted to fix the discrepancy.


leesjensen (approved):
Thanks!

PR #154, opened 2024-11-02
Instruction improvements

frozenfrank:
See the detailed commit messages for a description of the changes proposed:
* db61d28 Extend code quality language improvements...
* c62e2ce Reformat and expand common Phase 4 issues...
* 1dda948 Wrap file path in backquotes to present as code...
* 5f4c6be Exclude Intelli-J's .idea folder in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+.idea/
--- a/chess/4-database/database.md
+++ b/chess/4-database/database.md
@@ -9,7 +9,7 @@ In this part of the Chess project, you will create a MySQL implementation of you
 1. Install the MYSQL database management system (DBMS) on your development machine.
 1. Modify `db.properties` to contain your username and password.
 1. Design your database tables (i.e., your database `schema`)
-1. Implement a MySQL implementation of your Data Access Interface. Initially you can just stub out all of the methods.
+1. Implement a MySQL implementation of your Data Access Interface. Initially you can just stub out all the methods.
 1. Add the ability to create your database and tables, if they don't exist, when your server starts up.
 1. Iteratively write a test for each of your Data Access interface methods along with the backing MySQL code.
 1. Ensure that all provided pass off tests work properly, including the DatabaseTests added for this assignment, and the StandardAPITests from the previous assignment.
@@ -57,7 +57,7 @@ void storeUserPassword(String username, String clearTextPassword) {
 }
 ```
 
-Then when a user attempts to login, repeat the hashing process on the user supplied login password and then compare the resulting hash to the previously stored hash of the original password. If the two hashes match then you know the supplied password is correct.
+Then when a user attempts to log in, repeat the hashing process on the user supplied login password and then compare the resulting hash to the previously stored hash of the original password. If the two hashes match then you know the supplied password is correct.
 
 ```java
 boolean verifyUser(String username, String providedClearTextPassword) {
@@ -167,7 +167,7 @@ On the AutoGrader, SQL table names and SQL column names are **case-sensitive**. 
 
 ### Pass Off, Submission, and Grading
 
-All of the tests in your project must succeed in order to complete this phase.
+All the tests in your project must succeed in order to complete this phase.
 
 To pass off this assignment use the course [auto-grading](https://cs240.click/) tool. If your code passes then your grade will automatically be entered in Canvas.
 
--- a/chess/4-database/database.md
+++ b/chess/4-database/database.md
@@ -108,7 +108,7 @@ As part of your unit test deliverable you need to meet the following requirement
 1. Ensure that all of your unit tests work, including the new DAO tests and the Service tests you wrote in the previous assignment.
 
 > [!IMPORTANT]
-> You must place your data access test cases in a folder named server/src/test/java/dataaccess.
+> You must place your data access test cases in a folder named `server/src/test/java/dataaccess`.
 
 ### Code Quality
 
--- a/chess/4-database/database.md
+++ b/chess/4-database/database.md
@@ -114,18 +114,69 @@ As part of your unit test deliverable you need to meet the following requirement
 
 For this phase the auto grader will grade the quality of your project's source code. The rubric used to evaluate code quality can be found here: [Rubric](../code-quality-rubric.md)
 
+### Common Submission Issues
+
+Since you're adding code that acts with an external program in Phase 4, the likelihood of having auto-grader test
+output different from your local output is higher than previous phases.
+
+If you encounter this, here are some common things to try (expand the collapsible section).
+
+<details>
+
+<summary>Tips for resolving AutoGrader submission issues when local tests pass</summary>
+
+> [!IMPORTANT]
+> This guide assumes that your tests **are passing locally**.\
+> If your tests are not passing on your own computer,\
+> then you should definitely not assume they will pass on the AutoGrader.
+
+#### Database and Table Creation
+
+It's possible your server doesn't create the database and tables correctly when starting up.
+
+Try the following steps to reproduce the problem on your local machine:
+1. Drop your database/schema using an external tool (MySQL shell or workbench).
+1. Rerun the tests locally.
+1. If they fail, double-check the code for creating the database and tables.
+
+#### Hardcoded Database Name
+
+It's fun to assume that your database will always be named "chess", but the AutoGrader will break that assumption!
+
+The auto-grader inserts a new `db.properties` file for grading with most of the values different from those in your file.
+
+Run through the following verifications in your code:
+- Check for any place you may have hardcoded any values from `db.properties`.
+- Check each SQL statement, including where you create tables, for the database name.
+   - For example, use `INSERT INTO table` instead of `INSERT INTO database.table`.
+- The `getConnection` method inside `DatabaseManager` already sets up the connection to use your database,
+  so you shouldn't need to specify the database name if you are using that method to obtain your connections.
+
+#### CasE SEnSiTiVitY
+
+> [!CAUTION]
+> **Windows** users are especially susceptible to this issue since
+> Windows machines behave differently than the AutoGrader.\
+> Read this section carefully.
+
+On the AutoGrader, SQL table names and SQL column names are **case-sensitive**.
+Some personal computers treat SQL names with case insensitivity;
+this leads to disparities when submitting your code.
+
+- Double-check the casing in all of your SQL statements.
+  - For example, if you have table `foobar`,
+  - your machine may accept `INSERT INTO FooBar`,
+  - but the auto-grader machine will not.
+- Make sure each table and column name uses the same casing each time you use it.
+
+</details>
+
+
 ### Pass Off, Submission, and Grading
 
 All of the tests in your project must succeed in order to complete this phase.
 
 To pass off this assignment use the course [auto-grading](https://cs240.click/) tool. If your code passes then your grade will automatically be entered in Canvas.
-
-If your tests are passing locally but not on the autograder, here are some things to try:
-
-- It's possible your server doesn't create the database and tables correctly when starting up. Drop your database/schema using an external tool (MySQL shell or workbench) and rerun the tests locally. If they fail, double check the code for creating the database and tables.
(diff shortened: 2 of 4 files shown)


chess/4-database/database.md
```diff
@@ -108,25 +108,80 @@ As part of your unit test deliverable you need to meet the following requirement
...
+Since you're adding code that acts with an external program in Phase 4, the likelihood of having auto-grader test
+output different from your local output is higher than previous phases.
+
+If you encounter this, here are some common things to try (expand the collapsible section).
+
+<details>
+
+<summary>Tips for resolving AutoGrader submission issues when local tests pass</summary>
+
+> [!IMPORTANT]
```
frozenfrank, line 132: This alert and the following do not render properly within the collapsed section. See this discussion for the reported issue: [GitHub Community discussion #143379](https://github.com/orgs/community/discussions/143379).

Any recommendations on the next best option?

chess/4-database/database.md
```diff
@@ -114,19 +114,70 @@ As part of your unit test deliverable you need to meet the following requirement
...
+
+#### CasE SEnSiTiVitY
+
+> [!CAUTION]
+> **Windows** users are especially susceptible to this issue since
+> Windows machines behave differently than the AutoGrader.\
+> Read this section carefully.
+
+On the AutoGrader, SQL table names and SQL column names are **case-sensitive**.
+Some personal computers treat SQL names with case insensitivity;
```
leesjensen, line 167: Would the be more clear if it used operating system rather than personal computer?

chess/4-database/database.md
```diff
@@ -114,19 +114,70 @@ As part of your unit test deliverable you need to meet the following requirement
...
+Run through the following verifications in your code:
+- Check for any place you may have hardcoded any values from `db.properties`.
+- Check each SQL statement, including where you create tables, for the database name.
+   - For example, use `INSERT INTO table` instead of `INSERT INTO database.table`.
+- The `getConnection` method inside `DatabaseManager` already sets up the connection to use your database,
+  so you shouldn't need to specify the database name if you are using that method to obtain your connections.
+
+#### CasE SEnSiTiVitY
+
+> [!CAUTION]
```
leesjensen, line 161: This seems redundant with the sentence below. Is it necessary to repeat this here? If it was removed would it make this section more readable?

chess/4-database/database.md
```diff
@@ -114,19 +114,70 @@ As part of your unit test deliverable you need to meet the following requirement
...
+> **Windows** users are especially susceptible to this issue since
+> Windows machines behave differently than the AutoGrader.\
+> Read this section carefully.
+
+On the AutoGrader, SQL table names and SQL column names are **case-sensitive**.
+Some personal computers treat SQL names with case insensitivity;
+this leads to disparities when submitting your code.
+
+- Double-check the casing in all of your SQL statements.
+  - For example, if you have table `foobar`,
```
leesjensen, line 171: Is this supposed to be a list?

chess/4-database/database.md
```diff
@@ -114,19 +114,70 @@ As part of your unit test deliverable you need to meet the following requirement
...
+  - your machine may accept `INSERT INTO FooBar`,
```
leesjensen, line 172: By introducing `INSERT INTO` when talking about case it makes it seem like the capitalization of the command is the important thing. I almost missed that the `FooBar` case had changed.

chess/4-database/database.md
```diff
@@ -114,19 +114,70 @@ As part of your unit test deliverable you need to meet the following requirement
...
 For this phase the TAs will grade the quality of your project's source code. The rubric used to evaluate code quality can be found here: [Rubric](../code-quality-rubric.md)
 
+### Common Submission Issues
+
+Since you're adding code that acts with an external program in Phase 4, the likelihood of having auto-grader test
+output different from your local output is higher than previous phases.
+
+If you encounter this, here are some common things to try (expand the collapsible section).
+
+<details>
```
leesjensen, line 128: I have been trying to avoid using HTML syntax in MarkDown since it is not widely supported and would be complex to automatically transform for a different rendering.

chess/4-database/database.md
```diff
@@ -114,19 +114,70 @@ As part of your unit test deliverable you need to meet the following requirement
...
+
+<summary>Tips for resolving AutoGrader submission issues when local tests pass</summary>
+
+> [!IMPORTANT]
+> This guide assumes that your tests **are passing locally**.\
+> If your tests are not passing on your own computer,\
```
leesjensen, line 134: Why the line breaks?

chess/4-database/database.md
```diff
@@ -114,19 +114,70 @@ As part of your unit test deliverable you need to meet the following requirement
 
 For this phase the TAs will grade the quality of your project's source code. The rubric used to evaluate code quality can be found here: [Rubric](../code-quality-rubric.md)
 
+### Common Submission Issues
+
+Since you're adding code that acts with an external program in Phase 4, the likelihood of having auto-grader test
+output different from your local output is higher than previous phases.
```
leesjensen, line 124: Was the line break here intentional?

chess/3-web-api/web-api.md
```diff
@@ -422,7 +422,9 @@ Successfully run the service unit tests that you created. They must directly cal
 
 ### Code Quality
 
-For this phase the auto grader will grade the quality of all your project's source code. The rubric used to evaluate code quality can be found here: [Rubric](../code-quality-rubric.md). You can also test your quality with the auto grader independent of a specific phase submission.
+For this phase the auto grader will grade the quality of all your project's source code.
```
leesjensen, line 425: My editor wraps long lines, why make these explicit? This will make it more difficult to maintain the paragraph when changes are made.

.gitignore
```diff
@@ -1 +1,2 @@
-.DS_Store
\ No newline at end of file
+.DS_Store
+.idea/
```
leesjensen, line 2: Do we really want to do this? Pet shop is in the instruction repo and has a `.idea` folder that needs to be maintained.

chess/4-database/database.md
frozenfrank: Good point. That would be more clear.

chess/4-database/database.md
frozenfrank: I think the **CAUTION** alert helps identify the significance of the issue, and I agree that the wording could be cleaned up to reduce duplication between the two text bodies.

chess/4-database/database.md
frozenfrank: I prefer to have short readable statements, but you're right, this doesn't need to be a list.

chess/4-database/database.md
frozenfrank: There are really two core issues here, and I want to address them each individually:

## Markdown Html Support
I don't know what you're talking about when you say that "HTML syntax in MarkDown is not widely supported."

All of the implementations I've used of MarkDown support standard HTML tags by treating MarkDown as a _superset_ of HTML. One of the problems with Markdown since the early days has been the inconsistency in implementations of the language, and this has been addressed and largely resolved by the CommonMark specification (see [Why is CommonMark needed?](https://commonmark.org/#:~:text=Why%20is%20CommonMark%20needed%3F)).

There are non-standard things that would expectably break on a migration to a different system, but using HTML isn't one of them. (Those would be GitHub specific features like: Alerts, code linking, and other niceties.) We are largely sticking to standard MarkDown and HTML practices, and we're doing good by not implementing our own homegrown version of a styling language; that has happened in other departments on campus, and it quickly becomes a limiting factor.

For our foreseeable purposes:
* Intelli-J renders HTML (and this particular element)
* GitHub renders HTML (and this particular element)
* Most common markdown interpreters render HTML

Are there any key platforms which notably _do not_ support HTML tags inside markdown?

## This particular `<details>` element
The HTML `<details>` element is a standard HTML tag labelled as "Baseline: widely available" by the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details). It is used to contain information in a collapsed-by-default state which users can then click to request additional information. This pattern is specifically documented and recommended by GitHub in an article titled [Organizing information with collapsed sections](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections). The pattern is also supported by [GitLab](https://docs.gitlab.com/ee/user/markdown.html#collapsible-section).

In this document, the `<details>` tag is used to wrap around a pretty beefy (and growing) section of debugging tips which common visitors to the page will not need. This improves the overall readability of the page by allowing students to reach the end of the document with the rubric and important notes without getting slogged down by the technical debugging details. Students who are currently facing an issue will be more than happy to open up the collapsed container and read about the solutions to their problems.

Alternative approaches to this organizational choice include:
* Leave the content flatly on the document and increase the overall length of the file
* Extract the debugging tips to a separate page where they can be isolated
* Remove the debugging tips and expect students to just figure it out

All considered, I feel like the collapsible section is a great blend of the best of each of these approaches. It gives students the information they need when they need it, but it avoid overburdening them with the details when they don't need it.

I do not feel like the use of HTML tags will hurt the portability of the project. The tag itself is a widely supported tag, and the feature is well documented, and multiple common flavors of markdown including the ones we use, support it.

## Screenshots
### Intelli-J
#### Collapsed (default)
image

#### Expanded
image


### GitHub
#### Collapsed (default)
image

#### Expanded
image

chess/4-database/database.md
frozenfrank: I personally like to have short sentences on their own line to emphasize their importance. I like to control the way the text wraps so that each sentence is squarely on it's own line and the note forms a rectangular block rather than a wide line.

This is the writer in my coming out to express a preference about the way the text displays on the screen (something Steve Jobs wouldn't complain about). However, I am in now way going to say that my preferences are the best ones. I'm not claiming to be a UI designer, but I can do my best.

This is a detail that can be easily accommodated if you prefer that I don't exert control on the line wraps. Just reply back if you want them gone.

chess/4-database/database.md
frozenfrank: My editor didn't have softwrapping turned on. Now I have and will undo these explicit line breaks.

chess/3-web-api/web-api.md
frozenfrank: Heard. I turned on line wrapping by default for `*.txt` and `*.md` files in my editor, and I'll put the content all on one line.

.gitignore
frozenfrank: This has been addressed in commit 39ed688d24a0452360350df7f757ffe09501bd8f.

See the screenshot of the `git` system correctly recognizing new files in the `petshot/.idea/` folder (it already accepted changed files without this change).

image

chess/4-database/database.md
frozenfrank: I moved the content into a separate page to account for the concerns about using HTML.

chess/4-database/database.md
frozenfrank: This issue has been avoided by extracting the collapsible section into a separate page.

chess/4-database/database.md
frozenfrank: This section has been rewritten and expanded to highlight the important difference and provide additional information.

chess/4-database/database.md
frozenfrank: This has been consolidated.

frozenfrank:
Coming back to this after a long time, I found a resource that explicitly points out how Markdown syntax allows and permits HTML tags within the document. I'm not wanting to reverse the direction we took on this issue, but I just want to share what I found.

> For any markup that is not covered by Markdown’s syntax, you simply use HTML itself. There’s no need to preface it or delimit it to indicate that you’re switching from Markdown to HTML; you just use the tags.
> — [Markdown Syntax by John Gruber](https://daringfireball.net/projects/markdown/syntax#html:~:text=For%20any%20markup%20that%20is%20not%20covered%20by%20Markdown%E2%80%99s%20syntax%2C%20you%20simply%20use%20HTML%20itself.%20There%E2%80%99s%20no%20need%20to%20preface%20it%20or%20delimit%20it%20to%20indicate%20that%20you%E2%80%99re%20switching%20from%20Markdown%20to%20HTML%3B%20you%20just%20use%20the%20tags.)

PR #160, opened 2024-11-08
Add note about HTTP body responses for non-200 status codes

### LINKED ISSUE #157: Teach about HTTP Body Responses Coming from a different Stream on Error
As discussed in the TA meeting today, when the HTTP status code is not `200`, the body content cannot be read with in the normal way, but it must be read from an Error Stream instead.

ThanGerlek:
Closes #157 (unless we want to add it somewhere else as well?)

### DIFF
--- a/instruction/web-api/web-api.md
+++ b/instruction/web-api/web-api.md
@@ -256,6 +256,9 @@ public class ClientExample {
 }
 ```
 
+> [!TIP]
+> When the response code is not 200 (which can be checked with `http.getResponseCode() != HttpURLConnection.HTTP_OK`), you will need to use `http.getErrorStream()` to read the response body instead of `http.getInputStream()`.
+
 If you first run the `name list` service defined above, then you can run the `ClientExample` and see the full round trip HTTP request being handled by your Java code.
 
 ```sh


leesjensen (changes requested):
Does this need to be a tip? It seems more like an important part of the instruction. Something like having a second more complete example such as:


We can handle different HTTP response codes by checking the result from `http.getResponseCode()`. Additionally, you will need to read to get the response for the `getErrorStream` method if you receive a non-success status code. This is demonstrated in the following example:


```java
public class ClientAdvancedExample {
    public static void main(String[] args) throws Exception {
        URI uri = new URI("http://localhost:8080/error");
        HttpURLConnection http = (HttpURLConnection) uri.toURL().openConnection();
        http.setRequestMethod("GET");

        http.connect();

        // Handle bad HTTP status
       var status = http.getResponseCode();
        if ( status >= 200 && status < 300) {
            try (InputStream in = http.getInputStream()) {
                System.out.println(new Gson().fromJson(new InputStreamReader(in), Map.class));
            }
        } else {
            try (InputStream in = http.getErrorStream()) {
                System.out.println(new Gson().fromJson(new InputStreamReader(in), Map.class));
            }
        }
    }
}
```

leesjensen (approved):
Looks great

PR #161, opened 2024-11-08
Broaden "board orientation" requirement

ThanGerlek:
Now only requires the ability to demonstrate both orientations, rather than requiring them to both be printed together.


leesjensen:
It seems like we should just make it so that it displays according to the color that they choose to join as. I'm going to change the instruction to reflect that rather than given different options for how it can be done that need to be changed in later phases.

PR #162, opened 2024-11-08
fix(p5) to render as expected in p6

leesjensen:
This is an alternate solution to PR #161. Instead of allowing an arbitrary display for phase 5 we should make it look like phase 6 is expected to look.


frozenfrank (approved):
I don't know if you were looking for feedback from us TAs, but this looks great to me. I'm totally happy with this adjustment— it think it will only encourage students to do things the right way from the beginning.

Would it be helpful to recommend that students make the entire board work in one orientation first, before generalizing it to work in either orientation?

frozenfrank:
Note: I would love to help resolve the merge conflicts... but I don't have any write access to the repo yet ;) See #165

PR #163, opened 2024-11-12
Demonstrate Error Stream Reading in PetShop Example with Test

frozenfrank:
Regarding Issue #157, it was recently discovered that Spark requires the error output to be read from a completely different stream (`getErrorStream()`).

Upon inspection of the PetShop example code, I discovered that the example does not even send the detailed error messages over the internet. This leads to frustrating experiences when students cannot see the actual error messages being sent. The example `PetServer` code was adjusted to send detailed error messages, and the `ServerFacade` is adapted to read and throw the error messages.

Tests are included that prove the error serialization and throwing work as expected. The test expects an error to be thrown when attempting to add a dog with fleas. This endpoint was originally returning a `200` status code (which is not an error); the status code was updated to [418 I'm A Teapot](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418) for the following reasons:
1. Represent an error status code
2. Avoid conflicts with other known status codes
3. Hopefully help someone laugh somewhere along the way

### DIFF
--- a/petshop/server/src/main/server/PetServer.java
+++ b/petshop/server/src/main/server/PetServer.java
@@ -46,7 +46,7 @@ public class PetServer {
 
     private void exceptionHandler(ResponseException ex, Request req, Response res) {
         res.status(ex.StatusCode());
-        ErrorResponse errorResponse = new ErrorResponse(ex.getMessage());
+        ErrorResponse errorResponse = new ErrorResponse("Error: " + ex.getMessage());
         res.body(new Gson().toJson(errorResponse));
     }
 
--- a/petshop/server/src/main/server/PetServer.java
+++ b/petshop/server/src/main/server/PetServer.java
@@ -1,6 +1,7 @@
 package server;
 
 import com.google.gson.Gson;
+import exception.ErrorResponse;
 import exception.ResponseException;
 import model.Pet;
 import server.websocket.WebSocketHandler;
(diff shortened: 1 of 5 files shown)


frozenfrank:
@19mdavenport I know you can't approve this PR, but do you think this is a helpful adjustment to the PetShop example code will help students? The proposal for these changes originally came after banging _my_ head on the computer with a student for awhile until we figured out why a misleading error message was being thrown that didn't actually represent the error message the student threw.

frozenfrank:
@leesjensen Are you impressed with the test I included to prove that the functionality actually works? (I learned that from you; thanks!)

19mdavenport:
> Regarding Issue https://github.com/softwareconstruction240/softwareconstruction/issues/157, it was recently discovered that Spark requires the error output to be read from a completely different stream (getErrorStream()).

Pretty sure it's not spark, it's the HttpUrlConnection class.

frozenfrank:
> > Regarding Issue #157, it was recently discovered that Spark requires the error output to be read from a completely different stream (getErrorStream()).
> 
> Pretty sure it's not spark, it's the HttpUrlConnection class.

Thanks for pointing that out. In any case, I think we should have a code example that demonstrates proper usage of the classes.

19mdavenport:
> The proposal for these changes originally came after banging my head on the computer with a student for awhile until we figured out why a misleading error message was being thrown that didn't actually represent the error message the student threw.

Sounds like a skill issue /s

In all seriousness, the actual problem there is the student copied code they didn't understand. Overall, this seems like it would make it easier for students to directly copy and paste this code into their chess code without having to think about exactly what it's doing, which I am generally against, although having something to demonstrate using `getErrorStream` is probably a good thing. Ultimately up to the professors to decide if this is a positive change or not.

leesjensen (approved):
I think this is a useful addition. I'll probably tweak a few small things, but overall I like the contribution it makes.

PR #171, opened 2024-11-23
Remove indications that each feature is worth individual points...

### LINKED ISSUE #170: Phase 6: Consolidate Rubric Posted in GitHub
## Overview

In a recent TA meeting, we decided to consolidate the Phase 6 rubric to be all or nothing. The change to the Canvas Rubric needs to be mirrored in the instruction.

## Resources
* [Phase 6 GitHub rubric](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/6-gameplay/gameplay.md#grading-rubric)
* [Canvas Rubric](https://byu.instructure.com/courses/25927/assignments/951030)

## Screenshots

*Canvas*
image

frozenfrank:
Resolves #170.

## Implementation Details

Originally, I attempted to solve this by listing all of the features in a bulleted list in a single row of the table. This most accurately reflects the grading realities; however, it is very difficult to maintain since markdown files do not have builtin support for lists within tables. The workaround requires using HTML tags (which is fine), but then they all have to be condensed into a single line for the markdown to render properly.

In this approach, we remove the points from each item and indicate that they are sub-topics of the parent item which is 'Program Functionality'. I chose the abbreviation 'feat:' because it is short and conveys the intention of the full word 'functionality.'

I considered using ASCII characters to create a directory-like structure, but I figured that wouldn't actually work well from separate table rows.

### DIFF
--- a/chess/0-chess-moves/chess-moves.md
+++ b/chess/0-chess-moves/chess-moves.md
@@ -173,10 +173,10 @@ To pass off this assignment use the course [auto-grading](https://cs240.click/) 
 > You are required to commit to GitHub with every minor milestone. For example, after you successfully pass a test. This should result in a commit history that clearly details your work on this phase. If your Git history does not demonstrate your efforts then your submission may be rejected.
 
 | Category       | Criteria                                                                                          |       Points |
-| :------------- | :------------------------------------------------------------------------------------------------ | -----------: |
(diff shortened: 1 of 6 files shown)


frozenfrank:
I love that recommendation @leesjensen!

While I was at it, I also bolded the total rows for all the grading rubrics which is consistent with how a spreadsheet tool would automatically format a pivot table:

image

PR #172, opened 2024-11-25
Phase 6: Providing Ancillary Info about Run Configurations

### LINKED ISSUE #168: Phase 6: Provide Ancillary Info about Run Configurations
## Overview

Providing instructions on setting up multiple client configurations will help students with the particulars of using Intell-J.

## Description

There are multiple ways to accomplish this goal. We should make sure to use language that suggests possible approaches without eliminating any of the other approaches.

This is a topic that currently needs to be addressed every semester in a Slack message. Providing a persistent reference document would be very helpful.

This information could be added in a separate page, with only a link provided to it from the Phase 6 project page.

## Potential Approaches

Here are two approaches already shared by TAs in Slack:
* [Allowing multiple instances](https://cs240-fall2024.slack.com/archives/C07K66X928J/p1732203638383499)
* [Creating compound configurations](https://cs240-fall2024.slack.com/archives/C07JZHPK5N2/p1732295197013429)

Maillman:
This pull request resolves #168.
This pull request provides instructions on setting up multiple client configurations, required for passing off, which will help students with the particulars of using Intell-J. These instructions are on a separate markdown linked to in the Phase 6 project page. This pull request addresses a topic that previously needed to be addressed every semester in a Slack message.
If there are any issues with the wording or parts of the markdown that need to be clarified, please let me know.


leesjensen:
I added instruction in the debugging topic, along with a reference in P6, that details how to execute multiple instances. I really like the compound configuration ability, but I didn't include out of fear that students would be confused by it and is covered by the simple method of allowing multiple instances. 

fb03194ea77c6847cbf37137e67c89dd8842aca5

PR #222, opened 2024-12-10
fix(p6) rubric to reflect that all test cases must succeed

leesjensen:


frozenfrank (approved):
Looks good to me. Sorry for my oversight when I last edited this file.

PR #224, opened 2025-01-02
Use HttpClient instead of HttpURLConnection

### LINKED ISSUE #169: PetShop: Use HttpClient instead of HttpURLConnection
Currently, PetShop's `ServerFacade` uses `HttpURLConnection` to make requests. `HttpURLConnection`'s API is rather unintuitive and bulky, and JDK 11 provided `HttpClient` as a replacement.

Switching to `HttpClient` makes the requests simpler and more representative of a modern approach to http. It would also close #157, #160, and #163, as`HttpClient`'s api does not differentiate between the input stream and the error stream.

I'm happy to provide the code/docs changes! They're all relatively simple. I just wanted to make an issue before making a PR to give some heads up.

21aslade:
This PR changes the HTTP clients in this repository to use `HttpClient` instead of `HttpURLConnection`, per #169.

Specifically, it changes [GetExample.java](https://github.com/21aslade/cs240/tree/http-client/instruction/web-api/example-code/client-web-api/GetExample.java), [PostExample.java](https://github.com/21aslade/cs240/tree/http-client/instruction/web-api/example-code/client-web-api/PostExample.java), and the examples in [web-api.md](https://github.com/21aslade/cs240/tree/http-client/instruction/web-api/web-api.md). [example-code/old](https://github.com/21aslade/cs240/tree/http-client/instruction/web-api/example-code/old) is unchanged, as it isn't linked anywhere and doesn't seem to be maintained.

Note that because `HttpResponse` doesn't support an equivalent to `HttpURLConnection::getResponseMessage` for some reason, that functionality is removed from the curl clone :frowning: I don't think that's worth the cost of continuing to use `HttpURLConnection`, though, as `HttpClient` makes the code much simpler.

Fixes #169

### DIFF
--- a/instruction/web-api/example-code/client-web-api/GetExample.java
+++ b/instruction/web-api/example-code/client-web-api/GetExample.java
@@ -1,39 +1,40 @@
 import java.io.IOException;
-import java.io.InputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse.BodyHandlers;
+import java.time.Duration;
 
 public class GetExample {
+    private static final HttpClient client = HttpClient.newHttpClient();
 
-    public void doGet(String urlString) throws IOException {
-        URL url = new URL(urlString);
+    public static void main(String[] args) throws IOException, InterruptedException {
+        var example = doGet("https://example.com/");
+        System.out.println(example);
+    }
 
-        HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+    public static String doGet(String urlString) throws IOException, InterruptedException {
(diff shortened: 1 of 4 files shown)


leesjensen:
I think this is a good change, but it is going to require that the videos that reference the HTTP Client connection functionality get updated.  I will therefore leave this to @jerodw and @rodham for consideration.

jerodw:
I agree that this is a good change, but now that we have online sections starting in Winter, we will need to wait until we can update the videos. I think we can make the changes over the summer and have them ready to go along with the migration from SparkJava to Javelin for Fall 2025.

frozenfrank:
Since this was approved by the instructions, I have created a `fall2025` branch and changed the PR to merge into it. We can accumulate other changes in that branch and make it live at the appropriate time.

I'll merge this PR so it doesn't stay on our "open" list any longer.

PR #229, opened 2025-01-10
Update chess-github-repository.md to clarify deliverable instructions

pknjohns:
Clarified second step of instructions for submitting url to course autograder.

### DIFF
--- a/chess/chess-github-repository/chess-github-repository.md
+++ b/chess/chess-github-repository/chess-github-repository.md
@@ -85,7 +85,7 @@ Submit the URL of your repository to the [auto-grading](https://cs240.click/) to
 https://github.com/<youraccount>/chess
 ```
 
-After providing the URL to the AutoGrader, also pass off this "GitHub Repository" assignment with the course [auto-grading](https://cs240.click/) tool. If you created your repository correctly then your grade will automatically be entered in Canvas.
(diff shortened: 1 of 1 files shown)


frozenfrank:
@pknjohns Can you give us some additional context about why these changes are necessary? What about the original instructions confused you? Based on the confusion, what did you do incorrectly?

pknjohns:
Sure! When I first read the instructions, it sounded like it was telling me
to do the same thing twice (submit my github url to the autograder). I
didn’t realize until after I opened the autograder that you submit your
url, and then you have to assign it to an assignment. So I didn't end up
doing anything incorrectly based on the confusion, I just had to go into
the autograder itself to realize what the instructions were saying. Does
that make sense?

Thanks!

frozenfrank:
@pknjohns I just reworked the entire section to better address the confusion you experienced. Can you give it a review to see if the new version of the instructions would have been more clear to you?

I debated adding a screenshot of the completed process as well. Partly, I was deterred because any screenshot I generate now will show a late penalty. Is it important enough to try harder to get a good screenshot, or is the clarified wording good enough?

frozenfrank (approved):
Looks great to me! What else do the professors want before merging this in?

pknjohns:
That looks great, thank you! I think the clarified wording is good enough.

Best,
Parker

PR #230, opened 2025-01-14
Correct castling rules in chess-game.md

JansenNye:
Better wording of previous patch

### DIFF
--- a/chess/1-chess-game/chess-game.md
+++ b/chess/1-chess-game/chess-game.md
@@ -89,8 +89,7 @@ This is a special move where the King and a Rook move simultaneously. The castli
 
 1. Neither the King nor Rook have moved since the game started
 2. There are no pieces between the King and the Rook
-3. The King is not in Check
-4. Both your Rook and King will be safe after making the move (cannot be captured by any enemy pieces).
+3. The King is not in Check, does not cross a square on which it would be in Check, and is not in Check after castling.
 
 To Castle, the King moves 2 spaces towards the Rook, and the Rook "jumps" the king moving to the position next to and on the other side of the King. This is represented in a ChessMove as the king moving 2 spaces to the side.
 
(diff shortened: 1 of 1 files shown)


frozenfrank:
@19mdavenport This student has corrected the language in our instructions which previously had been incorrect.

After talking with the student in person, I'll point out some of the nuances that have been corrected:
* It doesn't matter if the rook is being threatened
* None of the squares the King crosses over are threatened

Are these statements already evaluated correctly in the [Castling tests](https://github.com/softwareconstruction240/chess/blob/main/starter-code/1-chess-game/passoff/chess/extracredit/CastlingTests.java)?

@JansenNye Can you also give a look over the castling tests and report/propose corrections if they are not properly asserting conditions consistent with the clarifications in this PR?

frozenfrank:
> @JansenNye Can you also give a look over the castling tests and report/propose corrections if they are not properly asserting conditions consistent with the clarifications in this PR?

@JansenNye Update: I just did that and submitted https://github.com/softwareconstruction240/chess/issues/1 which summarizes my findings. Feel free to comment on or submit a PR to resolve it if you like!

jerodw:
I agree with Michael. The castling rule is correct as previously stated. It's just stated in a confusing way. I think the new wording is better, so let's change it, but it wasn't incorrect before.

frozenfrank:
Now I see it. Thanks to https://github.com/softwareconstruction240/chess/issues/1#issuecomment-2591366279 I now see how the two statements are equivalent, and I do agree that the proposed statement is more clear in its intention and emphasis on king safety.

I just added another commit which updates and clarifies a few details.

@jerodw What other changes would you like to see before merging the PR? Are you OK with the bold formatting at the beginning of the line, or would you prefer a different presentation?

leesjensen (approved):
Thanks!

PR #237, opened 2025-01-23
Updating chess-moves.md Link to in-depth Equals and HashCode content

AldenKim:
As suggested during the TA meeting, we thought it would be helpful to include the link to documentation explaining why we actually need to override equals and hashCode functions.


frozenfrank (approved):
Looks good to me! I already pushed a few minor edits to the branch. Thanks for submitting this so quickly!

We'll still need a professor to actually merge this into the instruction repo.

PR #239, opened 2025-01-26
Update chess-game.md Adjust formatting...

frozenfrank:
- Use proper sub-headings rather than mere bold. This allows for directly linking to a subsection.
- Remove inappropriate use of backtics for non-code. Emphasize the terms Castling and En Passant by their location in the sub-headings.
- Add backtics around `ChessMove` code reference.

### DIFF
--- a/chess/1-chess-game/chess-game.md
+++ b/chess/1-chess-game/chess-game.md
@@ -79,11 +79,11 @@ By default, a new `ChessGame` represents an immediately playable board with the 
 
 ## Extra Credit Moves
 
-If you would like to fully implement the rules of chess you need to provide support for `En Passant` and `Castling`.
+If you would like to fully implement the rules of chess you need to provide support for "En Passant" and "Castling".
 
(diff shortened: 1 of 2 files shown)


chess/1-chess-game/chess-game.md
leesjensen, line 1: While it is true that things Like Castling or En Passant are not code, the idea was to just highlight the phrase. We do this all over in the documentation. We can change them to Bold or Italic if you the **code/command** highlight doesn't feel right to you, but I don't think putting them in quotes is the right thing to do.

frozenfrank: What if I turned them into blue links to the subsections? Would that feel appropriate to you?

frozenfrank: @leesjensen Upon further thought, the blue link idea doesn't help anyone. I replaced the quotes with bold in commit 6cad0a4.

Are there any other adjustments you would like to se?

leesjensen (approved):
Thanks!

PR #248, opened 2025-02-22
Add Example Error Case (Phase 2)

19mdavenport:
As discussed in the meeting yesterday, this PR adds an example of an error case to the register endpoint in the diagram and diagram discussion. It also adjusts the wording from "should" to "may" for including error cases.

If you want to view this rendered as markdown: https://github.com/softwareconstruction240/softwareconstruction/blob/add-phase2-example-error-case/chess/2-server-design/server-design.md

James also fixed a tip that previously wasn't showing correctly.


chess/2-server-design/server-design.md
```diff
@@ -123,6 +121,9 @@ This architecture includes a handler method for each server endpoint that calls
 
 You can decompose your handlers, services, and data access components into multiple classes or leave them as a single class as your design requires in order to meet the principles of good software design.
 
+> [!TIP]
+> You are not required to create your handlers in their own distinct classes. The [Web API instruction](../../instruction/web-api/web-api.md#implementing-endpoints) shows several other patterns, some of which are shorter and easier to use.
+
```
jerodw, line 123: Please remove the last phrase (some of which are shorter and easier to use). This biases students to thinking they shouldn't implement it the way it is shown in the diagrams, and there really are just pros and cons that are or can be discussed in lecture.

frozenfrank: Addressed in 19f048c

jerodw:
This looks good but please change it to merge into the next spring branch.

frozenfrank:
> This looks good but please change it to merge into the next spring branch.

@jerodw Done. We had previously discussed about this being unnecessary since it doesn't change any course requirements. At your request, the base branch has been changed.

jerodw:
This likely won't affect anyone currently in the course since we're past this phase, but I want to avoid changing specs in the middle of a semester if we don't have a critical need to change it.

frozenfrank:
> This likely won't affect anyone currently in the course since we're past this phase, but I want to avoid changing specs in the middle of a semester if we don't have a critical need to change it.

That makes sense. Thanks for sharing the thinking behind the decision. If we _really_ want to, we can send students a link to the file on this branch [devious face]

PR #250, opened 2025-02-25
Petshop sequence diagram

frozenfrank:
## Overview

Add a full PetShop sequence diagram as an example, and also provide a video overview showing the relationship between the diagram and the code.


petshop/server/src/main/server/PetServer.java
leesjensen, line 1: `/pet` is intentional. Similar to how a database table is not pluralized the endpoint path is not pluralized. The idea is that you want to to `delete /pet/*` not `delete /pets/*`.

petshop/shared/src/main/server/ServerFacade.java
leesjensen, line 1: TODO? Was this accidentally left in?

leesjensen:
This seems like fine work, but is it really necessary? Are the problems here significant and creating a stumbling block that degrades the students progress?

It does come at a cost. Keeping all of the content up to date (slides, github content, and videos) is expensive. I'm not sure this is worth doing.

petshop/shared/src/main/server/ServerFacade.java
frozenfrank: Yes, it's intentional. This is still a "draft" PR, and I needed to save the link somewhere before I closed the tab. It will be moved before I'm done.

petshop/server/src/main/server/PetServer.java
frozenfrank: Thanks for sharing your perspective on this.

## My Original Thinking

`delete /pet/*` seems like a nice approach, but the URI doesn't actually require the `*` character. This comes with the nasty _astonishment_ factor that accidentally calling this endpoint without a pet ID will unintentionally erase **all** of the pets.

## Online Research

I spent some time to look up common API naming conventions. Overall, it seems there are a variety of personal interpretations out there.

### Stack Overflow

[Should I use Singular or Plural name convention for REST resources?](https://stackoverflow.com/questions/6845772/should-i-use-singular-or-plural-name-convention-for-rest-resources) 

From the information I found, there are strong supporters who take different arguments. Some argue for consistency for programmatically generating the endpoints; others argue for linguistic correctness. Some argue that APIs should follow the same naming conventions as table/class names; others argue that APIs should not be subjected to the same rules as different kinds of concepts since there are additional considerations. Some compare API endpoints to folder names which can be singular; others point out that default folder structures are often plural in form. Some point out that naming conventions are arbitrary and ultimately hidden to the user; others defended the software developers who _do care_ about the kinds of things.

Among all the debate, the wise ones commented the following:
> "Neither way is right or wrong, go with what you like best.". Ah the famous line I hear so often and get sick and tired of hearing from people. Conventions matter and SHOULD be debated constructively amongst the community, that's where better solutions come about and good practices. — [PositiveGuy on StackOverflow (comment)](https://stackoverflow.com/a/6846057/2844859)

_A summarized representation of the arguments is following in a follow-up to this comment._

### Other Sources

[inmensoSofa Blog](https://inmensosofa.blogspot.com/):

> REST is not a protocol, but an application architecture style [[4]](http://en.wikipedia.org/wiki/Representational_state_transfer).

[Moesif Blog](https://www.moesif.com/blog/technical/api-development/The-Ultimate-Guide-to-REST-API-Naming-Convention/#:~:text=Plural%20vs%20Singular,by%20singular%20nouns.) (April 22, 2024):

> Plural nouns naturally represent collections or repositories of resources, making them the ideal choice for naming URIs. However, there is an exception to the rule. Singleton resources, which correspond to a single instance or specific record, are represented by singular nouns.

### Authoritative Source

Among all war of opinions, the only authoritative source I found is this Google API naming convention.

[API-122 Resource Names](https://google.aip.dev/122#collection-identifiers):

> Collection identifiers **must** be plural. (emphasis original)

## Conclusion

After seeing many different arguments, I think there are multiple valid perspectives on this issue, and that the most important thing is to be internally consistent while applying good principles as best as possible.

Personally, I think that API endpoints represent actions and that their naming should reflect the format of the data returned. This would fall on the linguistic side of the arguments. If an endpoint is `GET /user/:id`, then a singular user should be returned. However if the endpoint is `GET /users?groupId=...`, then I would expect an array of multiple items to be returned that match the query.

Specifically, I think it violates the Principle of Least Astonishment (POLA) to call `DELETE /pet` and find out that _all_ of the pets were deleted. As a programmer, I may have forgotten front-end validation to require a pet ID from a u, and now the entire collection is now wiped. This seems like a design flaw similar to that lead to UniSuper's 125 B private cloud being [accidentally deleted because of an omitted argument](https://www.datacenterdynamics.com/en/news/google-cloud-accidentally-deleted-unisupers-private-cloud-subscription/#:~:text=The%20company%20said%20of%20the,to%20leaving%20a%20parameter%20blank.)— bad design.

frozenfrank: I asked ChatGPT to summarize and group together the main ideas from the [StackOverflow question](https://stackoverflow.com/questions/6845772/should-i-use-singular-or-plural-name-convention-for-rest-resources) since the contributors represented so well a diversity of different perspectives. The question is currently closed as being "opinion-based."

_(Independent of the outcome, I'm amazed at AI's ability to process and summarize large amounts of information!)_

---

Here is a breakdown of the main arguments in favor of different naming conventions for HTTP API endpoints, along with direct quotes from the discussion grouped accordingly.

## **Plural Always**
**Key Arguments:**
- **Consistency**: All URLs start with the same prefix.
- **Logical Structure**: Collections should be plural, and IDs specify one resource.
- **Industry Standard**: Many public and private APIs use this convention.
- **Avoids Code Complexity**: Using both plural and singular complicates routing and logic.

**Quotes:**
1. *"Simple - all urls start with the same prefix. Logical - orders/ gets an index list of orders. Standard - Most widely adopted standard followed by the overwhelming majority of public and private APIs."* — [Eric Knudtson]
2. *"Ids specify ONE resource cars/id."* — [PositiveGuy]
3. *"When you are using both plural and singular for resource names in URIs, it complicates your code and the API because the user and the code behind the API has to account for that in routes and logic to differentiate single vs. plural whereas if you just stick with plural all the time you have no problems."* — [PositiveGuy]
4. *"Instead of thinking of it like 'get resource #123', phrase it in your head like 'get from the collection of resources that matches #123'."* — [Warren Rumak]

## **Singular Always**
**Key Arguments:**
- **Consistency with Code & Databases**: Classes, ORM tools, and database tables often use singular names.
- **Avoids Pluralization Issues**: Some words have irregular plurals or don't pluralize well (e.g., trousers, sunglasses).
- **Readability**: URLs might be more logical in certain cases.

**Quotes:**
1. *"Classes are always singular. ORM tools generate tables with the same names as class names. As more and more tools are being used, singular names are becoming a standard."* — [Sorter]
2. *"Consider this related resource: This /order/12/orderdetail/12 is more readable and logical than /orders/12/orderdetails/4."* — [Sorter]
3. *"For things without singular names (e.g., trousers, sunglasses), they appear to be singular by use."* — [Sorter]
4. *"This makes sense to me. However, we're a database-first shop, meaning we generate code and API entities from our database schema. And database standards tend to advocate singular table names, so we're going with that."* — [André C. Andersen]
5. *"I have always preferred to use Singular instead of plural. User.getList(), User.getById, User.delete etc."* — [Eastern Monk]

## **Plural for Collections, Singular for Entities**
**Key Arguments:**
- **Conceptual Clarity**: The plural represents the collection, while the singular represents an entity.
- **Natural Language Alignment**: A plural collection better reflects how we talk about lists and items.
- **Standardized Convention**: Many REST APIs follow this approach.

**Quotes:**
1. *"The premise of using /resources is that it is representing 'all' resources. If you do a GET /resources, you will likely return the entire collection. However, the individual resources are available at /resource/123."* — [Will Hartung]
2. *"If you read /employees/12 as 'the employee with id 12', how would you represent a subset? The only option is by making URIs more complex or distinguishing collections from the objects themselves (i.e., singular vs plural)."* — [Erik]
3. *"You can have a resource that is a collection of items and a resource that represents an item itself."* — [Erik]
4. *"To the advocates of singular, think of it this way: Would you ask someone for an order and expect one thing, or a list of things? So why would you expect a service to return a list of things when you type /order?"* — [Eric Knudtson]

## **Other Considerations**
**Arguments related to practical API usage, documentation, and client interaction:**
- **HATEOAS Perspective**: The structure of URIs shouldn't matter if the client follows links from the server.
- **Mapping to Code**: Plural naming simplifies how URLs map to collections in codebases.

**Quotes:**
1. *"The difficulty or ease of this is due to not respecting HATEOAS. It shouldn't matter whether it's plural or singular or anything else. You should respect the URI's sent from the server and not 'build up' your URIs on the client."* — [Richard]
2. *"The mapping also has the benefit of making documentation and tests on routes incredibly easy to write."* — [Thanos]

### **Summary of Perspectives:**
1. **Plural Always:** Simple, standard, avoids complexity in client and server logic.
2. **Singular Always:** Matches class and database naming, avoids pluralization confusion.
3. **Plural for Collections, Singular for Entities:** Clearer semantics for collections and individual items.
4. **Other Considerations:** HATEOAS compliance and ease of documentation/testing may affect the decision.

leesjensen: Love the summary and discussion on the subject. I have done a similar survey in the past and can see both sides. I did make an intentional choice in this case and that was my only point. Not to defend it as correct or incorrect. Honestly I think that perhaps I should have done `pets` but I don't it is that _wrong_ that it is worth changing.

I do wonder why some one would be astonished when they call `[DELETE] /pet` without providing **any additional parameters** for what specific pet they thought they were deleting. What would their expectation be? That it would randomly pick a pet to delete? That it would delete the last added pet? This seems especially true, since the proceeding endpoint, `[DELETE] /pet/:id` explicitly demonstrates how to delete a specific pet. Perhaps they would be equally astonished that `[DELETE] /pets` didn't delete a specific, yet unnamed, list of pets.

I guess I could change the endpoint to be more explicit `[DELETE] /pet/*` but that has other confusing semantics.

frozenfrank:
This PR is being withdrawn. In its place, proposals are written up to collect, discuss, and propose changes.
* See #252

PR #253, opened 2025-03-05
Introduce ChessGame Equality Comparison Requirements

frozenfrank:
This supports the new ChessGame tests already approved and being introduced in https://github.com/softwareconstruction240/chess/pull/19


jerodw:
This should be merged into the spring 2025 branch instead of main

frozenfrank:
That is correct. It has been updated 👌  The spring branch didn't exist when this PR was originally created.

PR #254, opened 2025-03-07
Move GSON Type Adapter instruction from SQL to JSON & Serialization

ryangladden:
Some GSON Type Adapter text was placed in the JDBC page and I think it belongs in JSON & Serialization. I took the Type Adapter section out of JDBC page and kind of smashed it into the existing JSON & Serialization page and removed a little bit of redundancy. 

Not sure what your preferred method of fixing this is, but I made a pull request mostly to draw attention to this.

### DIFF
--- a/instruction/db-jdbc/db-jdbc.md
+++ b/instruction/db-jdbc/db-jdbc.md
@@ -336,51 +336,6 @@ Collection<Pet> listPets(Connection conn) throws SQLException {
 }
 ```
 
-### Type Adapters
-
-Sometimes you need to deserialize JSON data into a field that is defined as, or contains, an Interface. When that happens, you must register an adapter to tell Gson what concrete class it should use when deserializing the JSON.
-
-We can demonstrate that my changing our Pet record to contain a `FriendList` instead of a `String[]`.
-
-```java
-interface FriendList {
-    FriendList add(String friend);
-}
-
-record Pet(String name, String type, FriendList friends) {}
-```
-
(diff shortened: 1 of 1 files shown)


frozenfrank:
@ryangladden Thanks for contributing this! We'll look it over and send it along to the professors to review.

I can't promise their response, but this seems like a good idea to me. Thanks for going the extra mile to submit a PR!

leesjensen:
You make a good point about putting all the serialization code in one place and the `JSON & Serialization` topic seems like the right place. However, I think some things were mixed up in the changes you made. The `JSON` topic is now referencing a `FriendList` object which is not defined. It also duplicates much of the information in the Chess example that is provided below the newly inserted instruction. 

Perhaps it would be better to just drop the serialization topic, or have a link to the topic, from the `DB-JDBC` instruction.

ryangladden:
If I remember right, I think I had done very light skimming on this change and _mostly_ just did some quick copy and paste action. I'll take a look at this and try and do an adjustment of better quality (if needed)

leesjensen:
No worries, as I am sure you are busy wrapping up the semester. I can take what you started and finish it up.

TheDavSmasher:
@ryangladden If you plan on coming back to this, mention me whenever you do and we will do our best to improve the instructions

leesjensen:
I simply removed the discussion of type adapters from the db code since it is sufficiently covered in the json topic.

dc2c17ae7f97c759a76aba83649f5acbb36ecb3f

Issue #252, opened 2025-03-04
Give PetShop Some Love


frozenfrank:
## Overview

This issue is a place to propose improvements to the PetShop example code!

During an in-person discussion with @leesjensen, it was discussed that many improvements could be made to the PetShop example code. However, we also decided that it's best to merely collect the proposed adjustments and allow the faculty members to make them at the appropriate time.

## Initial Proposals

> [!NOTE]
> Other TAs, please comment or add sub-issues to contribute other proposed ideas!

~It may be appropriate to split some of these into separate (sub)issues which GitHub now supports. However, I'll provide them in one list for consolidation of presentation.~

This list has now been split into sub issues because it was growing too long and unruly to manage. The sub-issues are small and individually focused. They allow for focused discussion on particular items; they also allow for clear communication as some proposals are accepted or rejected independent of other proposals. 

See the children of this issue for more information.

### 1-8. Existing Proposals

These have been moved into sub-issues.

List item 5 (listPets return type) was combined with list item 4 (request & response classes).

### Others...

Accepting additional feedback and collaboration. Please comment or create new issues.

IsaacGubelin:
I also support the idea of updating petshop. Not necessarily to make it _exactly_ like chess, but just like the above mentioned ideas. Var is nice, but I agree it creates ambiguity for students looking at the code for the first time.

Model classes would be great. Even a single pet model example would help a lot of students see the purpose of using java records.

Maillman:
I also agree with most of the points James brings up, and especially the point brought up by Isaac about using the `var`. This course uses Java, which is a strongly typed language. If we intend to use Petshop as a reference, I think we should enforce explicit typing in its code and show students that, in Java, we care about our types.

I also think we should definitely remove the behavior that DELETE `/pet` clears all the pets in the database, especially if someone intends to delete a pet using a CURL command, but accidentally hits enter before they enter go to enter the id; and now, they have accidentally deleted all pets because of a simple misinput. It's also clearer that the call to DELETE `/pets` deletes all the pets instead of `/pet`.

frozenfrank:
> This course uses Java, which is a strongly typed language. If we intend to use Petshop as a reference, I think we should enforce explicit typing in its code and show students that, in Java, we care about our types.

@Maillman Thank you for bringing this up! (@mewilker has also discussed this passionately.)

[...]

PR #257, opened 2025-03-09
Petshop-sequence-diagram-only

frozenfrank:
Resolves #253.


leesjensen:
Love the example diagram. How do you envision the learner discovering the diagram?

The textual content seems like it is more about how to use sequence diagrams with a petshop example instead of documentation for pet shop. Does this make more sense in the instruction or in petshop?

frozenfrank:
@leesjensen 

> Love the example diagram. How do you envision the learner discovering the diagram?

I'm not done yet. This is still a _draft PR_ because it's in progress. Would you prefer if I wait to create a PR until it's all the way complete?

> The textual content seems like it is more about how to use sequence diagrams with a petshop example instead of documentation for pet shop. Does this make more sense in the instruction or in petshop?

I would refer to an existing instruction module/page on sequence diagrams, but I'm not aware of any in the course. I also don't feel like it's my place or responsibility to create one. Therefore, I have provided a brief introduction to sequence diagrams in general here; it addresses concerns I've heard from students and other pieces of feedback I've received while testing the diagrams with other technically minded people.

In fact, that is probably why we TAs face so many basic questions about sequence diagrams— there is no introduction in the current materials!

If you feel it would be better placed in a new instruction module, feel free to push commits to this branch (or `main`) directly.

frozenfrank:
> The textual content seems like it is more about how to use sequence diagrams with a petshop example instead of documentation for pet shop. Does this make more sense in the instruction or in petshop?

@leesjensen In commit ebedd47, the tone and perspective of the article where adjusted to focus on _teaching a new developer about the architecture in this particular system_. The introductory statements have been reworded to emphasize the connection between generic diagrams and the specific way the diagrams are being used in this document. This still leaves open the door for additional instruction elsewhere, but hopefully makes the statements meaningful in their own right in this document.

frozenfrank:
I just reviewed the class diagram and sequence diagram with @jerodw. He approved the correctness of the diagram, and made the following notes:
1. The diagram uses more aggregation than normal, although each aggregation arrow is justified in its intention
2. The Mermaid diagram layout is ugly: the lines are curved and crossing, lines pass underneath boxes, the directed dependency arrow should be dashed (not solid).
3. The overall structure of the diagram is not well arranged visually

I'll take the feedback and use a better tool to render the diagram in a more beautiful way.

frozenfrank:
@jerodw The diagrams have been translated to PlantUML. I was even able to change the setting to use standard `+`/`-`/`#` symbols for accessibility modifiers, which is a big plus!

I will submit a separate PR to illustrate the WebSocket interactions.

leesjensen:
I can see that perhaps PlanetUML gives more control over the layout of the diagrams, but the reason I chose Mermaid was that it is supported by GitHub. This makes it easy to just put the "diagram source" right in the markdown file. This eliminates the maintenance overhead required to generate, capture, and insert images in some other solution.

leesjensen:
I'm also concerned that this detailed level of architecture documentation is basically duplicating the code in a visual format. It will significantly increase the maintenance of the project. Want to make a simple change the parameters of a function?

  1. change the code
  2. open up PlanetUML (overhead)
  3. modify the diagram  (overhead)
  4. capture an image  (overhead)
  5. insert the image (overhead)
  
This feels similar to adding detailed comments that simply reproduce what the code is already saying. I would prefer a model where we provided high level documentation/images to get the big picture and then let the actual code provide the detailed documentation.

This not only helps remove maintenance overhead, but also teaches the learner:

1. that the code is the only real source of truth
2. to develop their ability to quickly read code
3. to write/value code that documents itself

> Code never lies, comments sometimes do.
> - Ron Jeffries

> The proper use of comments is to compensate for our failure to express ourselves in code.
> - Uncle Bob

frozenfrank:
@leesjensen @softwareconstruction240/professors 

Thank you for raising your concerns about the maintenance of diagrams in the CS 240 class materials, particularly for the PetShop project. I understand the importance of keeping documentation aligned with the codebase and the challenges that come with it. Below, I’ll address your points and explain why these diagrams are valuable, while also proposing ways to manage the maintenance burden effectively.

## Addressing Maintenance Concerns
I recognize that updating documentation, such as PlantUML diagrams, after code changes can feel like a burden. However, this is a necessary part of maintaining high-quality teaching resources, much like updating call signatures, tests, or dependent modules when a function parameter changes in professional software engineering. While minor, backwards-compatible code updates might allow documentation to lag slightly, significant changes require updates to ensure accuracy. That said, the PetShop code isn’t being updated frequently—several proposed changes have accumulated (see #252)—which reduces the immediate maintenance load. To further minimize this burden, we’ve included clickable links to online editors in diagram comments, provided comprehensive update checklists, and leveraged PlantUML’s “save as PNG” feature for easy file replacement.

## The Purpose of Diagrams in CS 240
The CS 240 class has evolved as it has grown, shifting its focus from simplifying code edits to teaching core software engineering concepts, especially in the PetShop portion of the course. These concepts—system architecture, class relationships, and function call flows—are critical for students new to large codebases. The diagrams were created thoughtfully, from a TA’s perspective, to address common student misunderstandings and gaps not fully covered by existing resources like professor lectures or JavaDoc. They target areas where students need extra support, enhancing their learning experience.

## Prioritizing Accessibility Over Ease of Editing
Our priority is accessibility, not ease of editing, aligning with BYU’s approach of catering to diverse learners. The diagrams, alongside videos and written descriptions, complement the code by presenting information in multiple formats. This replication (not duplication) of content from JavaDoc, instructions, and class videos creates a more robust learning environment. For many students still developing their ability to read code, these visualizations bridge the gap, making course principles more approachable and engaging.

## Teaching Students to Read Code
A core goal of CS 240 is to teach students how to navigate and understand codebases. The diagrams support this by visualizing the larger system and its processes. Walking students through the relationship between diagrams and code not only clarifies the PetShop project but also builds their skills in interpreting code—skills they’ll carry forward. In Phase 2, we ask students to produce sequence diagrams with call signatures, so it’s consistent for us to include this level of detail in PetShop’s sequence and class diagrams. This detail enables precise learning conversations about function calls and layer roles, though improvements in conveying conceptual differences are being discussed (see #280).

## PlantUML vs. Mermaid
I agree that Mermaid is easy to maintain and powerful—I originally created the diagrams in Mermaid for those reasons. However, its output didn’t meet the quality threshold needed for teaching materials. PlantUML was chosen for its clearer, more professional diagrams, better suited to educational purposes. Whether Mermaid is “good enough” is subjective, and I’d welcome a discussion with the other professors if there’s a strong preference for it. Regardless, the maintenance burden with PlantUML has been minimized as much as possible, and the improved student experience justifies the effort.

## Code as the Truth?
The argument that “code is the truth” doesn’t fully align with past experiences. Last semester, TAs improved the code, yet some professors continued referencing outdated information because they weren’t consulting the updated “truth.” This suggests that students and instructors rely on multiple resources—code, documentation, lectures, and diagrams. While code changes were proposed before the documentation was created (see #250) and rejected, leading to separate updates, this reinforces that the code alone isn’t always the sole source of truth in practice.

## Distributing the Workload
The current model, where all changes and demonstration code flow through one professor, isn’t sustainable given CS 240’s scale—multiple professors, a large TA team, and hundreds of students. This single point of truth becomes a single point of failure and overburdened with maintenance tasks. A more distributed approach is needed. I propose mobilizing the TA team to actively contribute to maintaining and updating course resources, including documentation. This would lighten the load, improve efficiency, and better serve the broader community relying on these materials.

## Conclusion
In summary, while maintaining diagrams requires effort, their value in making CS 240 more accessible and effective for students outweighs the cost. They address misunderstandings, teach code-reading skills, and support diverse learning styles, all while aligning with course expectations. By distributing maintenance responsibilities among TAs and leveraging tools to streamline updates, we can ensure these resources remain sustainable and impactful. I’d be happy to discuss further adjustments, including the PlantUML vs. Mermaid debate, to find the best path forward.

Thank you again for your input—I look forward to your thoughts!

leesjensen:
Concepts represented by this pull request were implemented in [5038ad](5038ad1f0257414f50ac64fd707e9a62c0c3ecb4) of the `next-fall-2025` branch.

PR #258, opened 2025-03-12
Update debugging-tips.md in Phase 4

ZakkeryDaRebel:
Including information of Altering Tables with DROP and a link to the GitHub page about altering tables

### DIFF
--- a/chess/4-database/debugging-tips.md
+++ b/chess/4-database/debugging-tips.md
@@ -6,17 +6,18 @@ Since Phase 4 requires interacting with an external program (the MySQL database)
 ## If your code doesn't pass on your own computer
 
 > [!IMPORTANT]
-> Your code should be able to create the database and tables correctly when the server is run. This is done with the `CREATE __ IF NOT EXISTS __` section of the code.
+> Your code should be able to create the database and tables correctly when the server is run from the tests. This is done with the `CREATE __ IF NOT EXISTS __` section of the code somewhere (such as in Server, not in Main since the tests don't interact with Main at all).
 
 #### Database and Table Creation
 
-Make sure that your code has the create database and table statements somewhere, and make sure that they are called before any method tries to access the database. For example, if someone is going to register a user, all the databases should be created before hand, or else when it tries to check the database to see if the username has already been taken, it will throw an error since you have no database.
+Make sure that your code has the create database and each of the create table statements somewhere, and make sure that they are called before any method tries to access the database. For example, if someone is going to register a user, the database and all the tables should be created beforehand, or else when the code tries to check the database to see if the username has already been taken, it will throw an error since you have no database.
 
-Also make sure that your create statements are in correct SQL Format. A good example can be found in [Pet Shop](https://github.com/softwareconstruction240/softwareconstruction/blob/main/petshop/server/src/main/dataaccess/MySqlDataAccess.java).
+Also make sure that your create statements are in correct SQL Format. A good example can be found in the [db-sql](https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/db-sql/db-sql.md) instructions.
 
 #### Altering Tables
 
 - Since the tables are created at startup, if you change your create statement to store `game` as LONGTEXT instead of VARCHAR(100), this update will not be seen until you either `ALTER TABLE` or `DROP TABLE`
+  - If you have set up MySQL shell or workbench, you can use these alter or drop commands from there instead of just writing it directly into your chess code.
   - Alter Table allows you to change your table, such as to add a new column
   - Drop Table will completely delete your whole table, completely deleting all data currently on it, which would allow you to recreate it with your `CREATE TABLE` statement, which can include your newly updated column variable types
   - This [Git Hub Link](https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/db-sql/db-sql.md#altering-tables) contains a bit more explanation and some examples, not only of altering tables, but other important information about mySQL Tables.
@@ -42,14 +43,13 @@ Try the following steps to reproduce the problem on your local machine:
 
 It's fun to assume that your database will always be named "chess", but the AutoGrader will break that assumption!
 
-The auto-grader inserts a new `db.properties` file for grading with most of the values different from those in your file. (For example, instead of calling the database chess, it will call it chess123456)
+The auto-grader inserts a new `db.properties` file for grading with most of the values different from those in your file. (For example, instead of calling the database chess, it may call it chess123456)
 
 Run through the following verifications in your code:
 - Check for any place you may have hardcoded any values from `db.properties`.
 - Check each SQL statement, including where you create tables, for the database name.
   - For example, use `INSERT INTO table` instead of `INSERT INTO database.table`.
-- The `getConnection` method inside `DatabaseManager` already sets up the connection to use your database,
-  so you shouldn't need to specify the database name if you are using that method to obtain your connections.
+- The `getConnection` method inside `DatabaseManager` already sets up the connection to use your database, so you shouldn't need to specify the database name if you are using that method to obtain your connections.
 
 #### CasE SEnSiTiVitY
 
--- a/chess/4-database/debugging-tips.md
+++ b/chess/4-database/debugging-tips.md
@@ -10,9 +10,9 @@ Since Phase 4 requires interacting with an external program (the MySQL database)
 
 #### Database and Table Creation
 
-Make sure that your code has these statements somewhere, and make sure that they are called before any method tries to access the database. For example, if someone is going to register a user, all the databases should be created before hand, or else when it tries to check the database to see if the username has already been taken, it will throw an error since you have no database.
+Make sure that your code has the create database and table statements somewhere, and make sure that they are called before any method tries to access the database. For example, if someone is going to register a user, all the databases should be created before hand, or else when it tries to check the database to see if the username has already been taken, it will throw an error since you have no database.
 
-Also make sure that your create statements are in correct SQL Format. A good example can be found in Pet Shop.
+Also make sure that your create statements are in correct SQL Format. A good example can be found in [Pet Shop](https://github.com/softwareconstruction240/softwareconstruction/blob/main/petshop/server/src/main/dataaccess/MySqlDataAccess.java).
 
 #### Altering Tables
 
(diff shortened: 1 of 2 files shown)


19mdavenport (changes requested):
I'm not sure of the best way to fix this, but this "debugging-tips" page is currently only about things to try if there's problems with the autograder. Either this added info should go on a different page (the specs or a new debugging-tips page-we could rename this page autograder-debugging-tips or something if needed), or the first few lines of this page should be altered, as well as the usage of this page in the spec

Maillman:
As the markdown currently stands right now, I would suggest renaming the page to something like "autograder-debugging-tips" as the beginning refers to the guide only being helpful if they are already passing the tests locally but not on the AutoGrader: https://github.com/softwareconstruction240/softwareconstruction/blob/b05ac109fadc2354b526c2536bb332a40c8754ea/chess/4-database/debugging-tips.md?plain=1#L5-L6

frozenfrank:
I think we should adjust the language and make it clear that this page has
tips for multiple cases. I think we should break the page into sections:
1. Tests are NOT passing locally, AND
2. Tests are passing locally, but not in autograder

frozenfrank (changes requested):
I don’t like the rename. I don’t think this file should **always** be only about autograder problems. In fact, the whole point of this PR was to point out common problems that happen while still developing locally.

Rather than scoping this file down and creating a separate file for issues that happen locally, I think we should undo the rename and create new sections in this document.

frozenfrank:
> @frozenfrank
> 
> What sections on this tips page is only for is the Tests aren't passing locally. I thought that with the 'Important' section specifying that all these tests rely on the fact that the tests are passing locally but not in the autograder, that this whole section will apply to your 2nd point, not the 1st point.

Altering tables with DROP is a local development issue. If students never do this, they code would stop working on their local computer, but they would still pass on the autograder.

the sections do not yet exist. The warning was originally placed at the top of the entire file, but now that things are changing it think some headings (organization) should be added to indicate that it only applies to some of the tips, but not all of them.

ZakkeryDaRebel:
I'll work on this more next week. I'll add 2 sections, one for if the code works locally, but not on the autograder (Hardcoding, Case Sensitivity and Altering Tables). Then for the section for code that doesn't work locally, it will include Database and Table creation, including Drop.

webecke (approved):
I am a fan of this change

ZakkeryDaRebel:
Made some changes, but I was curious if anyone has any suggestions on what else we should add. The first section definitely can be approved with more information and rewording, but just curious what everyone thinks what information should be added to clarify any section.

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+## If your code doesn't pass on your own computer
+
 > [!IMPORTANT]
-> This guide assumes that your tests **are passing locally**. If your tests are not passing on your own computer, then you should definitely not assume they will pass on the AutoGrader.
+> Your code should be able to create the database and tables correctly when the server is run. This is done with the `CREATE __ IF NOT EXISTS __` section of the code. 
 
 #### Database and Table Creation
 
-It's possible your server doesn't create the database and tables correctly when starting up.
+Make sure that your code has these statements somewhere, and make sure that they are called before any method tries to access the database. For example, if someone is going to register a user, all the databases should be created before hand, or else when it tries to check the database to see if the username has already been taken, it will throw an error since you have no database.
```
Maillman, line 13: I would clarify what "these statements" refer to. Which I believe in this case is the create database and table statements.

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+
+Also make sure that your create statements are in correct SQL Format. A good example can be found in Pet Shop.
```
Maillman, line 15: Hmm, it may be a good idea to link to the Pet Shop code where the example is shown.

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+
+#### Altering Tables
+
+- Since the tables are created at startup, if you change your create statement to store `game` as LONGTEXT instead of VARCHAR(100), this update will not be seen until you either `ALTER TABLE` or `DROP TABLE`
+  - Alter Table allows you to change your table, such as to add a new column
+  - Drop Table will completely delete your whole table, completely deleting all data currently on it, which would allow you to recreate it with your `CREATE TABLE` statement, which can include your newly updated column variable types
+  - This [Git Hub Link](https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/db-sql/db-sql.md#altering-tables) contains a bit more explanation and some examples, not only of altering tables, but other important information about mySQL Tables.
+
+After you have altered your table, then you should be able to rerun your code, and check your database to see if the update was successful.
```
Maillman, line 24: I would adjust the first sentence to say "After you have altered **or dropped** your table..."

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+  - This [Git Hub Link](https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/db-sql/db-sql.md#altering-tables) contains a bit more explanation and some examples, not only of altering tables, but other important information about mySQL Tables.
+
+After you have altered your table, then you should be able to rerun your code, and check your database to see if the update was successful. 
+
+
+## If your code does pass on your own computer but not on the AutoGrader
+
+#### Database and Table Creation
+
+If your local tests are passing but the autograder isn't passing, there might be something different about your database and table creation than you might assume
```
Maillman, line 31: I'm not the biggest fan of how this sentence is worded, although I'm unsure how it would be worded more clearly at the moment.

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+
+Since your code will only create the database and tables if it isn't created already, it might be possible that you changed the code sometime while your database and table was still created, so it never actually updated your local SQL database.
```
Maillman, line 33: I would also reword this sentence to say something like "it might be possible that you changed the code sometime **after** the database and tables were created".

Maillman (changes requested):
The changes look pretty good so far, I don't know of anything else I would add to the tips already present. I just think some of the wording could be changed to be a little clearer, which I've already specified in other comments.

Maillman (approved):
I don't know what else could be added to the tips already present, but as far as I can tell, these changes looks pretty good to me!

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
 #### Database and Table Creation
 
-It's possible your server doesn't create the database and tables correctly when starting up.
+Make sure that your code has the create database and table statements somewhere, and make sure that they are called before any method tries to access the database. For example, if someone is going to register a user, all the databases should be created before hand, or else when it tries to check the database to see if the username has already been taken, it will throw an error since you have no database.
+
+Also make sure that your create statements are in correct SQL Format. A good example can be found in [Pet Shop](https://github.com/softwareconstruction240/softwareconstruction/blob/main/petshop/server/src/main/dataaccess/MySqlDataAccess.java).
+
+#### Altering Tables
+
+- Since the tables are created at startup, if you change your create statement to store `game` as LONGTEXT instead of VARCHAR(100), this update will not be seen until you either `ALTER TABLE` or `DROP TABLE`
```
19mdavenport, line 19: Might be worthwhile saying something along the lines of "You do not have to write code to do these operations. These can be done with an external tool (MySQL shell or workbench)". I've helped a couple of students over slack that have thought that us telling them to drop tables meant they needed to write the drop into code

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+## If your code doesn't pass on your own computer
+
 > [!IMPORTANT]
-> This guide assumes that your tests **are passing locally**. If your tests are not passing on your own computer, then you should definitely not assume they will pass on the AutoGrader.
+> Your code should be able to create the database and tables correctly when the server is run. This is done with the `CREATE __ IF NOT EXISTS __` section of the code. 
 
 #### Database and Table Creation
 
-It's possible your server doesn't create the database and tables correctly when starting up.
+Make sure that your code has the create database and table statements somewhere, and make sure that they are called before any method tries to access the database. For example, if someone is going to register a user, all the databases should be created before hand, or else when it tries to check the database to see if the username has already been taken, it will throw an error since you have no database.
```
19mdavenport: Might be worthwhile mentioning it needs to happen when the server is run from the tests, which uses the Server class directly rather than Main. I saw a few students this semester who had issues with that.

19mdavenport: "beforehand" is one word

19mdavenport: > all the databases should be created before hand

Usually we suggest having only one database and several tables. Ideally all the tables should also be created before any data access methods are called. I do, however, like your use of `should`, as I have seen successful (but inefficient) implementations call create table at the beginning of every single method in a DAO, so the table isn't created before the first method call but by the first method call.

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+
+Also make sure that your create statements are in correct SQL Format. A good example can be found in [Pet Shop](https://github.com/softwareconstruction240/softwareconstruction/blob/main/petshop/server/src/main/dataaccess/MySqlDataAccess.java).
```
19mdavenport, line 15: Not necessarily requesting this changed, but I would recommend linking the instruction (https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/db-sql/db-sql.md) rather than or before petshop. Too many people copy and paste those without understanding what any of `ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci` or what indexes do (which petshop is kind of weird with anyway)

chess/4-database/debugging-tips.md
```diff
@@ -2,23 +2,47 @@
...
+2. Rerun the tests locally.
+3. If they fail, double-check the code for creating the database and tables.
+
 
 #### Hardcoded Database Name
 
 It's fun to assume that your database will always be named "chess", but the AutoGrader will break that assumption!
 
-The auto-grader inserts a new `db.properties` file for grading with most of the values different from those in your file.
+The auto-grader inserts a new `db.properties` file for grading with most of the values different from those in your file. (For example, instead of calling the database chess, it will call it chess123456)
```
19mdavenport, line 45: I would suggest `may` rather than `will`

19mdavenport (dismissed):

leesjensen:
Looks great! Thanks.

PR #259, opened 2025-03-13
Update database.md

AldenKim:
Had a problem with a couple of student's not installing the MySQL server somehow, added the link to the instruction page to install MySQL at the top.


frozenfrank (approved):
Looks good to me :) Thanks for this simple adjustment. 

Do you want to link directly to a subsection of the document?

AldenKim:
> Looks good to me :) Thanks for this simple adjustment.
> 
> Do you want to link directly to a subsection of the document?

I don't think we ned to link to a subsection. This should be sufficient in my opinion.

PR #277, opened 2025-04-10
Fix Phase 3&6 Getting Started Tyrus Scope & Dependency Instruction

19mdavenport:
Adjusts the getting started for phase 3 to show using the `modules` tab instead of the `libraries` tab. [preview](https://github.com/softwareconstruction240/softwareconstruction/blob/tyrus-scope/chess/3-web-api/getting-started.md#dependencies) 
Adjusts the getting started for phase 6 to specify the scope for each module. [preview](https://github.com/softwareconstruction240/softwareconstruction/blob/tyrus-scope/chess/6-gameplay/getting-started.md#dependencies)


Maillman (approved):
These look like some pretty good changes!

PR #282, opened 2025-05-05
Linked the Javadoc webpage in the Programming Exam instructions

mewilker:


Maillman (approved):
Looks good!

TheDavSmasher (approved):
Looks good to me too!

leesjensen (commented):
I don't know if we specifically recommend. I couldn't find anything in the instructions, but it seems like we want to refer to the docs for a version greater than 8. 12 has string blocks and switch expressions. Records were introduced in 18 and so they might need the docs for that.

Perhaps 24 would be the best recommendation?

https://docs.oracle.com/en/java/javase/24/

mewilker:
I can definitely change that. The current link was what was linked before the GitHub page so that's why I used an old one.

mewilker:
I searched a little deeper and found that the Autograder runs it's own tests on the pipeline using JDK 21. I'm not sure what's the latest JDK it supports though. @TheDavSmasher would you happen to know?

TheDavSmasher:
@mewilker I'm not entirely sure how to even check that at the moment

TheDavSmasher:
@Maillman opinion?

Maillman:
I'm unsure what JDK the autograder runs on. But for the programming exam, I don't know if it matters too much... I personally have always used JDK 21 ever since I took the class, but I don't believe there should be any drastic changes between 21 and 24. I also just checked the link the programming exam in Canvas links to and it links to 21: https://docs.oracle.com/en/java/javase/21/docs/api/index.html

mewilker:
I apologize, I didn't see a link on a recent canvas version of the course! Guess I have fallen into the trap of not reading...

leesjensen (approved):
Thanks!

PR #283, opened 2025-05-07
Updated link and added period

lostinthework:
Added a missing period and fixed an incorrect link to the User-Agent Switcher and Manager Chrome extension.


mewilker (approved):
Nicely done! The link is a nice improvement because it takes you to the chrome store where you can download the extension.

Maillman (approved):
Seems like a really solid change here!

PR #286, opened 2025-05-14
Javalin Enable Automatic Pings Instruction

mewilker:
Spark's implementation of WebSocket would keep connections open for 5 minutes. This meant students generally didn't ping their clients to make sure connections were still alive. 

Now that we are migrating to Javalin, the connection timeout is 30 seconds. Thankfully, Javalin makes it super easy to ping it's connections. All you have to do is invoke the `enableAutomaticPings` function on the `context` object passed in. 

Currently, the instructions have a diagram referencing ping/pong, but don't describe how to enable it. @jerodw I know you said you put this code in your slides, but I couldn't find it on the ones linked here. I think it would be valuable to put this in the instructions somewhere. This pull request puts the `enableAutomaticPings` call in the sample echo server. It also has a section right after stating the value of it. 

If you feel like something is missing or inaccurate, please let me know!

Resolves https://github.com/softwareconstruction240/softwareconstruction/issues/285

### DIFF
--- a/instruction/websocket/example-code/server/SimpleWsEchoServer.java
+++ b/instruction/websocket/example-code/server/SimpleWsEchoServer.java
@@ -5,7 +5,10 @@ public class SimpleWsEchoServer {
         Javalin.create()
                 .get("/echo/{msg}", ctx -> ctx.result("HTTP response: " + ctx.pathParam("msg")))
                 .ws("/ws", ws -> {
-                    ws.onConnect(_ -> System.out.println("Websocket connected"));
+                    ws.onConnect(ctx -> {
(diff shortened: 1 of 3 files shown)


jerodw:
Thanks Makenna. This is great. Please add enableAutomaticPings to the WsRequestHandler code example too. I thought I had added this to the Javalin version of the WebSocket slides, but it looks like I didn't. I'll add it there. @rodham have you already recorded the updated Websocket lecture video? This would be a good thing to discuss there and point out in the slides.

Maillman (approved):
This looks good. Nice work, Makenna!

PR #289, opened 2025-05-15
/connect mysql shell

### LINKED ISSUE #287: MySQL Shell Setup: Connect While Using mysqlsh
## Overview
Some students, when they download the MySQL Shell, may click an option that already launches the shell for them but won't be connected to any MySQL server. This is also the case when someone simply types out 'mysqlsh' in the command prompt. However, the command provided in the [MySQL Markdown](https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/mysql/mysql.md) will not work for them and may leave them confused.
We should also probably provide instructions for students who are in the MySQL Shell but not connected to any MySQL server.

mewilker:
A student came in the other day, having launched mysql shell from the installation window. She was not connected to her server and came into the queue for help. This instruction shows students how to connect the shell to the server in such a case.

Resolves #287

### DIFF
--- a/instruction/mysql/mysql.md
+++ b/instruction/mysql/mysql.md
@@ -30,7 +30,9 @@ For example, if you created your root user with the password `edgarcobb` you wou
 mysqlsh -u root -pedgarcobb --sql
 ```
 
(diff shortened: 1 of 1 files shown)


Maillman (approved):
Pretty good change! I like it.

PR #292, opened 2025-06-05
Phase 5 Clarification

### LINKED ISSUE #270: Phase 5 Observer Functionality Problems
While trying to help people with passoffs, I had 2-3 people complain that the specs weren't clear enough on what they were supposed to do for Observe, seeing the "Functionality won't be implemented until Phase 6" section and just not have any code for it. 

I propose that instead of saying "Functionality won't be implemented", we should say, "Functionality for Phase 5 will just verify the game number passed in is a valid option, then print the board in White's perspective. Further Functionality will be implemented in Phase 6."

This will explain the two things we check for, that it verifies the number passed in is valid, and then prints out the board. 

What ideas does everyone have? Did anyone else run into similar issues with students? How do you think we can rephrase it better to limit confusion if that's even needed?

### LINKED ISSUE #271: Phase 5 Gameplay UI Leave
If students just print the board, and can still access post login menu (list, create, etc) then this doesn't apply to them. 

But if students want to enter a Gameplay UI, they should be able to print the board, but they also need to have the ability to `leave` the Gameplay UI, and return back to the post login menu. 

Because the specs doesn't specify the requirement to leave the Gameplay UI, some students have to restart the whole client process in order to continue. 

At the TA meeting we discussed that we do want this functionality, if they implement a Gameplay UI, though they don't have to actually implement a Gameplay UI and can just print the board instead. 

Pros:
- Testing will be easier without constantly restarting the client
- This is more professional and makes more sense to be able to not get stuck in your code

Cons:
- Since functionality will be implemented in Phase 6, students don't need to worry about gameplay UI functionality
- This change will only affect some students, as some will just print the board and not worry

mewilker:
I tagged as low change to the requirements because the language in this PR would enable the TAs to give a reason to reject some Phase 5 implementations. These changes should get students closer to what the professors expect from their programs.

If a student got into a gameplay state and couldn't get out, it would go against the statement that the program should not crash, freeze or lock. It doesn't explicitly say that an in-game menu should have some sort of way to get out of it, which is why I chose the verbiage. Shout out to Josh Clancy for helping me refine that. This should resolve #271

I also changed `Additional functionality will be added in phase 6`  to indicate that we would like a stub of observe to be present, which should resolve #270 and hopefully eliminate some confusion about what the observe command is supposed to do in phase 5.


Maillman (approved):
Hmm... I do wonder if a student may be more confused by how their client will "freeze" or "lock", but honestly, I can't figure out a way to word it any better. I think the changes are reasonable enough, however.

TheDavSmasher:
I might suggest something like explaining that if your program needs to be rerun to return to a previous state, that that counts as a lock. Since we're trying to include softlocks, not just hardlocks, in that restriction?

Also, when should this be merged? Fall? Or Summer? @mewilker

mewilker:
I removed the lock verb, and instead added a small section about getting stuck in a menu. I think this could go into summer, and maybe it wouldn't be a big deal if it got merged in right now, but I'll leave that up to @jerodw

jerodw:
This looks good to me now and I don't mind merging it now. Change it to a real PR and I'll merge it.

PR #298, opened 2025-07-03
File consistency

TheDavSmasher:
I found some inconsitencies with how some files were formatted (such as having the links before the title or having a H2 header instead of H1), which this PR addresses.


leesjensen (approved):
Thanks for removing all the absolute links.

Issue #297, opened 2025-07-01
Programing Exam Set Up Video


ZakkeryDaRebel:
One student from the Spring 2025 term felt that he was a bit unprepared for the Programming Exam and had wished that there was a video tutorial of what the flow of the Programming Exam would be like. For example this could include a video of someone starting a Programming Exam, following through the Honorlock steps to get it set up, and show how to upload their code and a screenshot to the exam. 

This does raise the question that if we want to do this, should we just have 1 video, or should we try to have a different video for each different OS (Mac and Windows for example). 

In order for a TA to take the test, they just need to enter Student mode to be able to access the test. That is if the professors think it's best for us TAs to handle this. 

This was the conversation between the Student and the TA:

`Student: test went smooth for me! glad i took the practice exam twice to know how the flow would go! wish there was a better tutorial on how the test flow would actually go but i guess you just want us to figure that out. `
`TA: Great work! Though I don't fully understand what you mean by wishing there was a better tutorial on how the test flow would be?`
`Student: Like just having a video of someone going thru the current honorlock process and submitting the code. Just feel like it might be helpful? I don’t know if that makes sense. I think it would have taken a bit of the ambiguity of the test itself. Just a thought`

ZakkeryDaRebel:
I'm going to close this issue because students have the practice programming exam to test with, so a video seems like a lot of unnecessary work.

PR #299, opened 2025-07-03
Update programming-exam.md

ZakkeryDaRebel:
Pull Request that links to issue #297 

Added some extra information about making sure students know how to zip a folder or take a screenshot, but maybe I should replace this with actual steps on how to zip a folder or take a screenshot?

I also added some extra information about Honorlock, specifically about being in a brightly lit room, and a quiet place. Not required, but I thought it could be a nice little line to add. 

Does anyone think additional information should be added, or how can me make sure people are prepared enough with the programming exam

### DIFF
--- a/instruction/programming-exam/programming-exam.md
+++ b/instruction/programming-exam/programming-exam.md
@@ -22,7 +22,8 @@ The purpose of this exam is really to show that you are able to tackle large and
 
 1. Do the exam setup:
     * Create a new clean project from starter code
-    * Take the HonorLock Setup/Acknowledgment Quiz 
+    * Take the HonorLock Setup/Acknowledgment Quiz
+    * Make sure you know how to zip a folder and take a screenshot
     * Take the Practice Programming Exam (optional)
 
 1. Review and practice. Make sure you have some time to review after Phase 0 and before taking the exam.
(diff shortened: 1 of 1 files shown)


mewilker (approved):
I generally feel like Honorlock proctors are pretty good at being reasonable with the students and telling them what they need to do to fix flags, so I don't know that there is anything else to mention about that. 

I personally feel that the knowing how zip a folder is pretty easy to look up, but maybe I need to have more charity. I'm ok without it but I can't imagine it would take too much time to add.

leesjensen:
These additions while completely reasonable, seem unnecessary to me. I think that we too quickly respond to one student's experience and keep adding more and more notes, details, and things to watch out for. This makes the instruction harder to read and actually results in students glossing over it because it is too long and complex. TLDR;

On another note.  The issue that was referenced #297 was about one student wanting us to create a tutorial. This pull request seems to focus on making sure the lighting is good and that you know how to make a zip file before you start. What is the disconnect here?

Regardless, both are useful, but I don't think either rise to the level of generally being helpful enough to justify the changes.

ZakkeryDaRebel:
Thanks for your thoughts @leesjensen 

We had talked about this issue in our meeting last week, and we were uncertain if we should create a video that may be difficult for maintaining if we change our policies or if we stop using honorlock. Ultimately we decided that we were just going to add some information, which I felt like I did my best adding information that would help people out with the test. 

I recognize that the programming exam information is already quite long and complex, so maybe my 2 little note changes about making sure students know how to zip a file and take screenshots may not have been helpful at all. If you feel like making a video would be really beneficial, then I can film a video and we can discuss if that is going to help people and not become too difficult to maintain. 

As for the notes about honorlock, it does have some disconnect, I felt like it could be helpful to add since as I have been grading some students, there have been some struggles with lighting and a lot of voice recognition, so I thought adding that could be helpful, but it is on a completely different topic than what was discussed in issue 297. I only included it because it is still important and could help people not feel so unprepared for honorlock issues (if they read it and follow it that is).

ZakkeryDaRebel:
This issue really would only help a small number of students (those with high anxiety or struggle to understand test flow), so we had discussed that out of all the students this semester, it may only help at most 3. So this isn't a super important issue or pull request that we can just close if needed. 

We could also try to do the video instead of adding more text and debate if the video is worth it to help the small number of students that may feel unprepared.

PR #301, opened 2025-07-08
Change starter diagram colors

### LINKED ISSUE #300: Phase 2 Starter Diagram Bad Contrast
sequencediagram.org has changed how they render the alt box colors. Before they had some sort of opaqueness to them and the tag on the top left matched the color of the boxes. Now they are highly saturated colors (that are hard on the eyes) and some of the boxes have bad contrast with the text. If you look at the current starter diagram, you can see black text on a navy background, and it's so hard to read. I'm planning to fix this myself, but I'm making an issue of it so I don't forget.

mewilker:
Resolves #300.

Apparently, sequencediagram.org introduced a feature where the whole group box no longer has to be one color. Unfortunately, it ruined our entire scheme for our endpoint colors. I tried to replicate the original, but it seems there's no way to change the text color for the group label. I attempted using tags that the documentation says should work, but it didn't. So I just redid the colors and all the text in the labels are black. The contrast is decent enough.

@rodham Would we want to merge this change on the main branch instead? I know students are working on Phase 2 already but the color scheme is so atrocious that all my students submitting have changed theirs.


Maillman (approved):
This looks good, let's get it merged into main as soon as possible. But one thing I've noticed is that the branch @mewilker has requested to merge into (next-fall-2025) seems to have some formatting issues. It looks like someone was trying to resolve some merge conflicts but didn't remove some of the merge conflict text. @softwareconstruction240/professors, please look at https://github.com/softwareconstruction240/softwareconstruction/blob/next-fall-2025/chess/2-server-design/server-design.md in preview and remove everything that's causing weird formatting issues.

TheDavSmasher:
@mewilker Since phase 2 is past now, can we change the PR to merge to next-fall so we can all approve and review?

PR #302, opened 2025-07-09
Change starter diagram colors NOW

mewilker:
Apologies for the duplicate PR of #301, but this would change starter diagram colors for the current term, and the other makes sure the wonky formatting Melvin noticed gets fixed. If there's a better way to make sure this happens, I'll happily change.


ZakkeryDaRebel (approved):
All those colors look a lot better

Maillman (approved):
Looks great! @softwareconstruction240/professors, let's get this merged in ASAP.

PR #303, opened 2025-07-15
Script and GitHub Action to maintain a Wiki frontend

### LINKED ISSUE #294: Move instruction and other markdown files to Wiki
This isn't a requirement but I feel that using GitHub's Wiki functionality could be useful to provide quick access links, a navigation tab, as well as being far easier to find information.

@softwareconstruction240/core

TheDavSmasher:
The Script provided takes all markdown files and edits them such that they are fitting for being Wiki pages, maintaining all navigation links functional (or as functional as they'd be in the base repository in the first place), keeping all embeds, and referencing back to the actual repository when desired, such as links to code files or folders.
Closes #294 

Currently, all pages maintain their title as given by the markdown page's H1 header (except for the README, which becomes the home page). The only exception is the Getting Started pages as all of them are named pretty much the same with no variation and it would lead to name conflicts, so the script currently appends " - Phase #" to the Getting Started pages to keep them unique and functional but the links to and from them still work.

This also provides a static "Sidebar" for navigation within the wiki. This is a markdown page stored within the .github folder as it's only needed for the Wiki. The current Sidebar (as of making this PR) is incomplete and has comments until I know what the professors want it to contain and in what order.

The script updates the wiki whenever there's a push to main, so it'll only change the wiki when the content the students would see changes.

The script is structured in such a way that to update any base cases or how it handles titles is all contained within a second file designed to hold this information so it's not necessary to edit the main script to maintain functionality if it were to change. Currently, there's no need to change it, but the option exists.

To work, the repo will need a single PAT with only repo access (only to this one, so fine-grained would work) stored as a repository secret, since the GitHub action relies on another built action that does the wiki update with the files the script prepares.

@softwareconstruction240/tas @softwareconstruction240/professors 

For a proof of it working, see [my fork of this repo](https://github.com/TheDavSmasher/softwareconstruction) and [its wiki](https://github.com/TheDavSmasher/softwareconstruction/wiki) with the same script.


TheDavSmasher:
Ignore how many commits I did, I tested in a very non-pleasant way so there's more commits than there would be otherwise. I'm fine stripping the code from my fork into here if desired.

TheDavSmasher:
Also, should this be merge to next-fall or main? I'll leave the question here for discussion, but there's still subtasks to do before it's "ready" ready

mewilker (approved):
This looks good to me!

ZakkeryDaRebel:
I think it looks really good. Schedule technically doesn't open the schedule, but goes to home where the students can click the schedule link. Was this done because the schedule updates every semester?

ZakkeryDaRebel:
As for when to merge it, I can see either immediately or next fall. Since it doesn't fundamentally change any requirements, we can merge it immediately, unless the professors would prefer it to be merged next fall to not have any changes.

Maillman (approved):
This looks pretty good! I appreciate the link to the help queue and the autograder on the sidebar as well.

rodham (approved):

PR #313, opened 2025-08-05
Link final exam review to Dr. Rodham's schedule

mewilker:


rodham:
I agree. I thought they were relative. I don't know how to absolute URLs got in there. I must have copied one of Jerod's schedules that had absolute URLs in it.

PR #317, opened 2025-08-23
add oversimplified diagram for en passant and castling

jclegg03:


leesjensen (commented):
Thanks for submitting this. It definitely is an area where students can get lost if they don't know the rules of chess. However, I think it is probably better to keep things really simple and point them to a better chess book for this type of instruction.

PR #327, opened 2025-10-01
Added comment on == vs .equals()

ZakkeryDaRebel:
I know a good amount of students try to do attackPosition == kingPosition and are confused why it will never return true, so I thought it would be a good idea to add that explanation here.


ZakkeryDaRebel:
I totally agree! I didn't see the `Java Object Class` instruction topic, so I had assumed that I need to explain. That is my bad, I should have done my research. I have simplified the paragraph to a simple fix and redirected students to that instructional topic so they can learn why it happens.

leesjensen (approved):
Thanks!

PR #328, opened 2025-10-03
Update chess-tips.md: added GitHub commit tips to each phase

bsharplydian:


ZakkeryDaRebel (approved):
This looks good to me! These are some good tips, and the advice to tell students to try and increase their commit count before coming in is really good as well!

leesjensen (commented):
Looks great.

PR #329, opened 2025-10-04
Add note for Drawing Board Video

ZakkeryDaRebel:
## Overview

This little addition notifies the students about the change that makes the video outdated and states what is actually expected, in case they look only at the video and not the rest of the specs.


leesjensen (approved):
Thanks

PR #331, opened 2025-10-15
Passoff examples

ZakkeryDaRebel:
# Overview

There are a lot of students for pass off days, and us TAs always get really busy, and students get really stressed because they want to make their one shot passing off with TAs count so they don't have to come back. This has led to some students requesting a video to be made to showcase what the passoff is going to be like. 

After discussing it in the TA meeting with the professors, we decided that a video would be unrealistic because it would require maintenance if we ever changed the requirements, and it could share too much information too the students. Instead, we found a middle ground to give a small maintainable list of suggestions the students should implement before passing off. 

## Benefits

Since there are a lot of students that come in, do part of the pass off, and fail, forcing them to come back and do either the whole pass off again or have TAs keep track of the students and where they failed so they can test only those failed requirements. By adding this passoff prep to Phase 5 and 6, students will be able to have a checklist of things that they should make sure works before coming in and passing off. This could limit the amount of times a student will need to come in for passoff, and help students feel more prepared for the pass off, instead of forgetting to implement a small handful of checks. 

## Discussion

Should anything be...
- reworded to clear confusion?
- added for students to also prepare for?
- removed to reduce the chance of students hard coding solutions instead of actually coding implementation?
- further discussed before we merge this into GitHub?

### DIFF
--- a/instruction/chess-tips/chess-tips.md
+++ b/instruction/chess-tips/chess-tips.md
@@ -235,7 +235,7 @@ Here are a couple of things that students commonly forget to include as part of 
 
 - After registering, you will automatically enter the signed-in state, you don't need to login afterwards. [Prelogin UI Command Descriptions](../../chess/5-pregame/pregame.md#prelogin-ui)
 - Make your ListGames numbering be independent of the game IDs. [Postlogin UI Command Descriptions](../../chess/5-pregame/pregame.md#postlogin-ui)
-- Make sure your board is printed properly! [Gameplay UI Description](../../chess/5-pregame/pregame.md#gameplay-ui)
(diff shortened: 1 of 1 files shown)


ZakkeryDaRebel:
Change to be common errors, instead of an exhaustive list that the students can complain about not being prepared for

ZakkeryDaRebel:
Put into TA Tips instead

ZakkeryDaRebel:
If there are any changes that should be made, let me know, otherwise I will make this a legit PR for the professors to merge.

leesjensen (approved):
thanks!

PR #347, opened 2026-01-08
Fix isGlobalOdd method to actually check for odd values

WyWyGuy:


leesjensen (approved):
Wow! That has been in there for a while. Thanks.

PR #357, opened 2026-02-20
New Phase 3 TA Tip

### LINKED ISSUE #356: Add TA tip phase 3 to clarify that Jackson dependencies will not work with the autograder
Students who try to serialize their data and send it back to the client in phase 3 using context.json() will find that their code will not compile. The TA tips/specs do not clearly specify that the jackson dependency is not used for this project and that serialization should be done with Gson. 

Image

TwoLettuce:
Closes #356 

Some students run into issues if they serialize their data using dependencies other than Gson, as those dependencies are not recognized by the autograder. This causes difficult issues for students when their code runs fine locally but won't compile or gives unexpected errors when run in the autograder, which is the reason I considered it requisite to add a TA tip to help students avoid this problem.

### DIFF
--- a/instruction/chess-tips/chess-tips.md
+++ b/instruction/chess-tips/chess-tips.md
@@ -100,17 +100,13 @@ Return new Gson().toJson(Map.of(“message”, ex.getMessage()));
 
 ## Autograder doesn't compile my project - package com.fasterxml.jackson.core does not exist / expected: <401> but was: <500>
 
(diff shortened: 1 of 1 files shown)


leesjensen (approved):
Isn't the root of the problem is that the student doesn't return the correct objects from their endpoint and so Javlin suggests jackson to handle the serialization. Perhaps a better, or additional, tip would be if you get the jackson suggestion from javalin, to actually check what you are returning.

Issue #295, opened 2025-07-01
DatabaseManager singleton makes testing difficult


leesjensen:
The global DatabaseManager class makes it difficult to do testing using dependency inversion. 

1. Properties are read from a file rather than providing them to the manager.
2. The singleton pattern makes it difficult to load different databases for different tests or to create and delete a database for a test.
3. It forces the same database to be used for both testing and development.
4. It forces [tests](https://github.com/softwareconstruction240/chess/blob/main/starter-code/4-database/passoff/server/DatabaseTests.java) to use introspection to implemented desired behavior at runtime.

Most implementations are already passing a DataAccess object around and so there is little value in a DatabaseManager singleton. We should just initialize the DataAccess object with the desired DatabaseManager. Then a test could provide a mocked database version if desired, or set up and tear down a test database.

leesjensen:
@rodham, @jerodw, @mewilker, @Maillman  - I did a quick test to see if this would make things better and it looks good.

**TL;DR;**

1. Code is much cleaner.
2. Gets rid of clear function (Jumping for joy.)
3. Gets rid of complex reflection calls in testing.
4. Allows for temporary testing databases that are deleted after tests. This keeps your dev database from getting polluted or cleared.

If we do this, here is the remaining work to do:

1. Replace starter code with new code. - Should only take a few minutes.
2. Replace instruction that references the old DatabaseManager. - Should only take an hour or so.
3. Fix references in videos if they exist? - this is the hard one and I don't know what it involves.

Except for what might be required for the videos, I could easily make this change next week after I merge the fall 2025 branch.

Here are all the details:

## DBManager

Created a `DBManager` interface with a `MySQLDBManager` implementation. 

```java
public interface DBManager {
    public void createDatabase() throws DataAccessException;
    public void clearDatabase() throws DataAccessException;
    public void deleteDatabase() throws DataAccessException;
    public Connection getConnection() throws DataAccessException;
    public String dbName();
}
```

`DBManager` needs to be interface in order to do dependency injection.

The `MySQLDBManager` class has two constructors. One that simply reads from the properties file, and one that allows you to override the database name. This keeps things compatible with how it was being used before but allows for **temporary** databases that can be created and destroyed during testing.

## Server impact

The server now takes a DBManager as a parameter. That is then passed to the DataAccess layer SQL implementation. This provides for the dependency injection.

```java
public class Main {
    public static void main(String[] args) throws Exception {
[...]

PR #360, opened 2026-03-19
Phase 4 error handling

mewilker:
# Explanation

The Phase 4 tests introduce a test to students that makes sure that when a `SQLException` or more appropriately a `DataAcessException` is thrown due to the inability to get a connection, the error is handled as an "Internal Server Error". While this test is good and teaches students to handle unforeseen errors appropriately, it references the Phase 3 spec. While it would be nice if this was implemented and Phase 3 and carried over seamlessly in Phase 4, my experience is that most students don't begin to error codes until this Phase 4. 

If their code works properly in Phase 3, they never really run into the 500 response. Sure, it would be nice for debugging if they caught their null pointer exceptions and sent a 500 back, but most of them don't do that. They typically either A) let Javalin execute its default behavior (catch the exception and print it to the console) or B) have an else statement for one of their status codes (like the 400 for bad request) and have that returned. They may struggle with this through Phase 3, but ultimately, they pass the `StandardAPITests` without ever having to worry about the 500 at the bottom of each of our endpoint tables.

Then comes along Phase 4. A lot of students seem confused about what the `Database Error Handling Test` means and spend some time trying to figure out what the test is asking for. Some of them spend some time trying to figure out what they are serializing and what exactly the test wants. Now I believe @Maillman's PR to change the tests should help clarify this (found here: softwareconstruction240/chess#49) but I think there also needs to be a reference to the Phase 3 spec in the Phase 4 one. This is so students can understand where this expectation comes from.  Rather than determine our expectations from the test, they are reminded in the spec how to handle database errors.

# Changes

This PR adds a small new section to the spec. The paragraph explains the response that is supposed to happen when connections occur. It also links to the Phase 3 spec, so they can be reminded of the format we expect.

# When to Merge:

This PR could be merged as soon as the professors deem it appropriate, however since Phase 4 has already passed, it may be a good idea to wait for Winter 2026 to end.

### DIFF
--- a/chess/4-database/database.md
+++ b/chess/4-database/database.md
@@ -78,10 +78,7 @@ The above code demonstrates the necessary concepts to implement secure password 
 
 ## Error Handling
 
-Now that your database is stored in MySQL, you must account for any errors trying to connect to that program. While your database is most likely kept on your own machine, it's possible that it could be kept on another. If the internet were to go down and the two programs are not able to establish a connection to each other, a `SQLException` would be thrown. 
-
-If this happens while an endpoint is trying to access the database, this should be considered an `Internal Server Error`, and an appropriate response must be handled.
-The status code must be set to `500` and the body of each of these responses must include a reasonable, relevant error message. You may recall this specification from [Phase 3](../3-web-api/web-api.md#endpoint-specifications).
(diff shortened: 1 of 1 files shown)


leesjensen:
Would it be better to implement [Issue 295](https://github.com/softwareconstruction240/softwareconstruction/issues/295) before we add more DB testing code. With #295 implemented it would be very easy to do failure testing.

PR #362, opened 2026-04-03
add: note to DataAccessException introduction.

bsharplydian:
This indicates that adding more Exception classes could be helpful for organization, but doesn't prescribe any details.


chess/3-web-api/web-api.md
```diff
@@ -190,6 +190,9 @@ void insertUser(UserData u) throws DataAccessException
 
 The starter code includes a `dataAccess.DataAccessException`. This exception should be thrown by data access methods that could fail. If a method call fails, it should throw a `DataAccessException`. For example, the `DataAccessException` is thrown if a user attempts to update a non-existent game. If you like, feel free to create subclasses of DataAccessException that represent more specific errors relating to data access.
 
+> [!NOTE]
```
leesjensen, line 193: Would it be helpful to give an example of why you might want to add other classes and possibly point to the [exception instruction](https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/exceptions/exceptions.md) rather than the note.

We might want to enhance the exception instruction to actually give an example in the `custom exception` section that shows how to extend the exception class and why you would do it.

bsharplydian:
Prof. Jensen, I added the changes you recommended. Let me know if there are any further changes you'd like me to make.

leesjensen:
Thanks

PR #364, opened 2026-05-05
Update Phase 0 Getting Started

mewilker:
Recent versions of IntelliJ have been displaying the root folder that the project is opened in. As long as this folder is not a module, the project works just fine. However, our instruction is written in such a way that students have been concerned about their builds when they are usually fine. 

This PR updates the wording, as well as adds an image so that it's clear that the chess folder being a module is the problem rather than the root folder.


leesjensen:
How do they get into a state where you cannot compile? What do they do about it when it happens?

I thought this was pretty straight forward. Clone the template repo. Open project in IntelliJ. Where are things going wrong?

chess/0-chess-moves/getting-started.md
mewilker, line 22: Usually it happens when they click the build with Maven prompt, displayed on line 22 of the Phase 0 Getting Started Markdonw. 

At the moment I am seeing a bunch of students with a correct setup but they come to the queue concerned that their setup is wrong because IntelliJ shows the chess folder. This change is supposed to assure them that having a chess folder is fine, so long as it is not a module in the program.

The way the documentation is currently worded is concerning students because the new versions of IntelliJ display the folder the project is opened in.

leesjensen:
Perhaps we can mitigate this by removing the maven scripts. Since the autograder doesn't use them, it seems like we could remove confusion by pulling them out.

Then we can just update the instruction to match the latest version of Intellij.

mewilker:
I can make that change if you want, but it would mean editing a few things. For one the Chess starter code. 

Another thing that would need to change is this section: https://github.com/softwareconstruction240/softwareconstruction/blob/main/instruction/aws-chess-server/aws-chess-server.md#build-the-client-and-server-jars

It goes over deploying the chess server and client. It is the only place I know of that we discuss how to build the client jar, and it uses the starter maven. There may be a video or some slides that go over it that I'm not familiar with. I think when I took the class there was a way to build the jar with Intellij, but I don't remember what it was. I could do some more research if you would like.

chess/0-chess-moves/getting-started.md
```diff
@@ -25,7 +25,11 @@ Your project should look like this when it opens with the client, server, and sh
 
 ![open intellij](open-intellij.png)
 
-There should not be a folder named chess showing up in your Intellij project, only the items inside of the chess folder should be visible. If you do see a chess folder, check that there are not two different folders named chess. You can confirm that the modules are set up correctly by going to File > Project Structure > Modules, then verifying only client, server, and shared are modules. Feel free to talk with the TAs for help.
```
mewilker, line 28: Regardless of what we decide this line needs to change, as there have been students coming in with their setup looking like the green example in the picture I'm proposing to add, and they are concerned that there is going to be a problem.

leesjensen:
Sorry for the long delay. I haven't forgotten about this. I am still trying to figure out how to make the overall process easier.

- I want to remove the pom file from the starter code. This is causing problems and isn't really necessary.
- I'm happy to change the AWS setup instruction or even **remove it entirely**. It is nice, but not really necessary. I would like to find a solution where they could publish their server, but the EC2 direction is just too complicated for the course.
- Change the instruction to match what the current version of intellij is providing. This should be clear instruction without things like "if it looks like this then do this, or looks like that then do this". It is critical that the initial setup be braindead easy. If it is not easy let's attack that rather than providing instruction that is complicated and requires careful study.

mewilker:
> * I'm happy to change the AWS setup instruction or even **remove it entirely**. It is nice, but not really necessary. I would like to find a solution where they could publish their server, but the EC2 direction is just too complicated for the course.


Publishing to EC2 does feel like a lot. I do think there needs to be a way to least build aritifacts. When I finished the class, I was able to send a client jar to my Dad we played chess on our home network.

I feel like if you can get it working on other friends computers, it's a good thing to show non-tech friends what you do as a CS major

mewilker:
Withdrawing so code coverage PR will have less to review

PR #367, opened 2026-06-03
Use Spring Schedule

mewilker:
The Winter schedule is up right now and spring is still going....


leesjensen:
FIXED 1df32bc55c1b2d0f4f9287c929cb14fc52a0967c

PR #371, opened 2026-07-25
Update prompt text in InteractiveCalculatorExample

efoulk4:


leesjensen (commented):
Thanks for the pull request. Can you help me understand why "input" is significantly better than "type"?

efoulk4:
Yes, I forgot that the PR's don't provide context for the whole file. Either is fine, I just suggested a change to "input" becasue that is what the sample output below displays. One could just as easily change the sample output to say "type"

leesjensen:
Ah that makes sense.

